Issue #6, opened 2021-04-23
poolpair.controller.ts


fuxingloh:
#### What would you like to be added:

Controller for poolpair.

```
GET v1/:network/poolpairs
GET v1/:network/poolpairs/:id
GET v1/:network/poolpairs/:id/swaps
GET v1/:network/poolpairs/:id/liquidity
GET v1/:network/poolpairs/:id/liquidity/add
GET v1/:network/poolpairs/:id/liquidity/remove
```

It should be designed to provide GET (stateless) pool pairs by the id. Poolpair information can be queried from the current attached node provided by `@src/module.defid`. 

The pool pair controller response data should not be a carbon copy of what is available in the RPC result. The model should be redefined to follow a semantic structure. Only what is required should be sent. This allows the evolution of this controller as we can use different source provider. As of now, we can piggyback on JsonRpcClient as the data source.

#### Semantic/Design structure we are using for DeFi Whale
- [RESTful API](https://en.wikipedia.org/wiki/Representational_state_transfer)
- RESTful resources naming semantic hence `/poolpairs` instead of `/poolpair`. Not all semantic has plural hence `/liquidity`
- camelCase following Javascript Style guide for the response body

PR #24, opened 2021-05-02
added poolpairs.controller.ts

canonbrother:

/kind feature

Implements poolpairs.controller.ts that exposes POST /v1/network/poolpairs for creating, getting and listing pool pairs, adding pool liquidity and listing pool shares


README.md
```diff
@@ -1,7 +1,7 @@
 [![CI](https://github.com/DeFiCh/whale/actions/workflows/ci.yml/badge.svg)](https://github.com/DeFiCh/whale/actions/workflows/ci.yml)
 [![codecov](https://codecov.io/gh/DeFiCh/whale/branch/main/graph/badge.svg?token=kBCC9qSRrA)](https://codecov.io/gh/DeFiCh/whale)
 [![Maintainability](https://api.codeclimate.com/v1/badges/593ffda9c1d91261a37b/maintainability)](https://codeclimate.com/github/DeFiCh/whale/maintainability)
-[![TS-Standard](https://badgen.net/badge/code%20style/ts-standard/blue?icon=typescript)](https://github.com/standard/ts-standard)
+[![TS-Standard](https://badgen.net/badge/code%20style/ts-standard)](https://github.com/standard/ts-standard)
```
fuxingloh, line 4: 🧐

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+   * @return {Promise<string>} hashes of string
+   */
+  @Post()
+  async create (@Body() body: PoolPairDto): Promise<string> {
+    try {
+      return await this.client.poolpair.createPoolPair(body.metadata, body?.utxos)
+    } catch (e) {
+      throw new BadRequestException(e)
+    }
+  }
```
fuxingloh, line 63: We don't need createPoolPair for DeFi whale as it will be running a stateless, wallet-less node where the only mutating function is it can submit raw txn. In another words, only GET operations.

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const options: PoolPairsOptions = {
+    isMineOnly: query.isMineOnly?.toLowerCase() !== 'false'
+  }
+
+  return {
+    pagination,
+    verbose,
+    options
+  }
```
fuxingloh, line 124: I believe https://github.com/typestack/class-transformer can do all these, which is part of this package.

You can remap Number, and e.g. can fail if `start` and `limit` is negative. You can make it strict and fail exceptionally if any validation fail.

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  @Post('/liquidity')
+  async addPoolLiquidity (@Body() body: PoolLiquidityDto): Promise<string> {
+    try {
+      const { from, shareAddress, options } = body
+      return await this.client.poolpair.addPoolLiquidity(from, shareAddress, options)
+    } catch (e) {
+      throw new BadRequestException()
+    }
+  }
```
fuxingloh, line 104: we don't need this too

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+    const filter = query !== undefined ? remap(query) : undefined
+    return await this.client.poolpair.listPoolPairs(filter?.pagination, filter?.verbose)
+  }
+
+  /**
+   * @param {string} symbol token's symbol
+   * @param {PoolPairsQuery} query pool pair filter
+   * @return {PoolPairResult}
+   */
+  @Get(':symbol')
```
fuxingloh, line 80: Best to add the prefix seperator in front for better clarity of code.

`@Get('/:symbol')`

fuxingloh (changes requested):
Requested an initial set of changes, will review other parts later.

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export interface CreatePoolPairMetadata {
+  tokenA: string
+  tokenB: string
+  commission: number
+  status: boolean
+  ownerAddress: string
+  customRewards?: string
+  pairSymbol?: string
+}
```
fuxingloh, line 106: We also need to follow a strict JSON in and out semantic for DeFi whale, since this is a REST service. I was looking at a [Stackoverflow](https://stackoverflow.com/questions/5543490/json-naming-convention-snake-case-camelcase-or-pascalcase) post comparing different types of naming convention. 

I think we can follow `snake_case` for all data transfer object (DTO, struct, POJO). An object that does not hold features but merely for use of transferring data between system. This allows consistency of thought where DTO class or interface have a very clean semantic. 

This will require the remapping of your JSON output, but I think this allows for cleaner implementations where consumer/producer contracts are clear and clean. Seeing as we will be owning the DeFi products (everything on top of blockchain) we can design a pattern of consistency into our core code.

> For reference https://github.com/DeFiCh/whale/blob/fuxingmodule/db-actor-address/src/module.database/models/block.ts (still WIP, this is part of a 3 module development I am working on now)

/cc @jingyi2811

fuxingloh: This also means you can look from a UI/UX building perspective and be selective about what data to transfer and not to.

You can remap DTO to fit a semantic that you think is clean for implementation.

fuxingloh: E.g. 
from
```ts
count_item: number
count_negative: number
```
to (as an example)
```ts
count: {
  item: number
  negative: number
}
```

fuxingloh:
As per the specs: https://github.com/DeFiCh/whale/issues/6

I think what I didn't mention was to provide a set of APIs for explorer/wallet/informatics to display relevant information. Bouncing ideas, not concrete but it should allow the UI implementation to just query and display without building complex query on the client-side. Hence to limit the query to a simple set of instruction, while the API service will perform them (complexly if required). Making it easier for dev usage.

#### Some thoughts:

GET v1/:network/poolpairs
- To get all poolpair with pagination

GET v1/:network/poolpairs/:id
- To get a specific poolpair

GET v1/:network/poolpairs/:id/swaps
- To see all the swaps in a pool pair with paginations

GET v1/:network/poolpairs/:id/liquidity
- To see all the add/remove liquidity in a pool pair
- To filter only add?
- To filter only remove?

GET v1/:network/poolpairs/:id/shares
- To see rich list?

src/main.ts
```diff
@@ -11,6 +12,7 @@ async function bootstrap (): Promise<void> {
     AppModule,
     newFastifyAdapter()
   )
+  app.useGlobalPipes(new ValidationPipe({ transform: true }))
```
fuxingloh, line 15: I believe this provide the same feature 

https://github.com/DeFiCh/whale/blob/c098a07c647753fa6000c9790c8adae9aedc6576/src/module.api/index.ts#L16-L22

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,302 @@
...
+  'reserveB/reserveA': BigNumber | string
+  tradeEnabled: boolean
+  ownerAddress: string
+  blockCommissionA: BigNumber
+  blockCommissionB: BigNumber
+  rewardPct: BigNumber
+  customRewards: BigNumber
+  creationTx: string
+  creationHeight: BigNumber
+}
```
fuxingloh, line 249: you could import your types definitions from `jellyfish-api-core`

For example:
```ts
import { poolpair } from '@defichain/jellyfish-api-core'
let name: poolpair.CreatePoolPairMetadata
```

fuxingloh (commented):
some consideration you might want to take note of

src/module.api/poolpairs.controller.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  /**
+   * List pool pairs with optional filters
+   *
+   * @param {PoolPairsFilter} [filter] filter of listing pool pairs
+   * @return {Promise<PoolPairInfoDto>}
+   */
+  @Get()
+  async list (@Query(new PoolPairsQueryPipe()) filter?: PoolPairsFilter): Promise<PoolPairInfoDto[]> {
```
fuxingloh, line 107: To follow the query specification in the recent pushed PR. Don't give the consumer the ability to query anything beyond one index. Lesser option is better.

fuxingloh (changes requested):

canonbrother:
close this PR and re-work with new structure design

PR #31, opened 2021-05-05
database module with an agnostic provider

fuxingloh:
/kind feature

### Database

DeFi whale uses a database-agnostic implementation. Any provider is valid as long as it can conform to the interfaces provided in the Database. Superset implementations such as RDMS are always supported.

For a performant design, the interface uses a log-structured merge-tree (LSM). LSM uses a log data structure with performance characteristics that allow for high write volume which is required for ledger-based application. This database interface implements a key-value structure with type, index, partition and sort as key.

### Models

> `Model`, `ModelIndex` and `ModelMapping` all together makes up the 3 points of contact for building the model in DeFi whale. 

`Model` describes the interface of the data with only `id` as a static requirement. 

`ModelIndex` defines what indexes are available in the LSM store. `ModelIndex` supports two types of indexes which are `ModelIndex` and `ModelIndexComposite`. 

- `ModelIndex` with just `partitionKey` represents a key-value structure where there is a single unique key in the index for each model. This supports your traditional queryById operations.
- `ModelIndex` with both `partitionKey` and `sortKey` represents a key-sort-value structure where there is a single unique composite key. The composite key being `partitionKey` and sortKey. Partition key as it suggests, partition the model index space into unique index slices. While the `sortKey`, sorts the unique index slices. Being a sorted key allows for sort operations to be performed on the sorted unique index slice. This supports your traditional queryById and gets 10 items starting from this sort key.

### Database Provider: LevelDB

> defich/ain uses LevelDB too.

LevelDatabase uses [Level/level](https://github.com/Level/level) with a LevelDB instance under the hood. [Level/subleveldown](https://github.com/Level/subleveldown) is used to divide key spaces into models are their partitions. Data stored in level are denormalized.

This is meant for production use but works well for testing too.

### Database Provider: Memory

MemoryDatabase uses [Level/memdown](https://github.com/Level/memdown). This is meant solely for testing.

### Conformation Testing & other database provider?

For both providers, conformation testing is done specification by specification to test that it can conform to the requirements for the DeFi whale database module.

LSM database offerings are incredibly commonplace, Apache Cassandra, Amazon DynamoDB, Google BigTable, Facebook RocksDB offers similar capabilities which allow for an agnostic implementation with a tiny surface level of 4 interfaces to implement to get it going. (RDMS is a superset and can conform to LSM specification. But won't be a performant)

### `module.models`

As an example, `src/module.models/block.ts` is mapped and added but is not targeted for feature completion in this PR.


src/module.database/provider.level/level.database.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  async put<M extends Model> (mapping: ModelMapping<M>, model: M): Promise<void> {
+    // TODO(fuxingloh): check before deleting for better performance
+    await this.delete(mapping, model.id)
+
+    const subRoot = this.subRoot(mapping)
+    await subRoot.put(model.id, model)
+
+    for (const index of Object.values(mapping.index)) {
```
ivan-zynesis, line 116: Interesting implementation, I wonder the possibility for not await for after insertion indexing for the main thread.

fuxingloh: for LSM database implementations it's usually done on a single thread (especially the embedded ones), running it async will still bottle it out during IO. This keeps consistency while there are no great performance benefits since we need all indexes to be written.

PR #37, opened 2021-05-07
added clean error handling

### LINKED ISSUE #22: clean error handling paradigm/semantics

We need a cleaner exception/error handling paradigm. Errors that exits the DeFi whale must have a clean serialization and allow deserialization on arrival on the consumer side. Semantics must be clean. Exceptions include validation, 500, bad request, gateway and etc.

#### Why is this needed:

Errors and exception like all things are part of the greater developer quality of life. This is important as it creates a clear and clean semantic for error handling across DeFi services. Error handling like all things, have to scale as our services scale.

An example of a similar spec I wrote a while back: https://github.com/fuxingloh/error-url

canonbrother:

/kind feature

Fixes #22


__e2e__/health.e2e.ts
```diff
@@ -69,8 +60,9 @@ describe('/_health/probes/readiness', () => {
           status: 'down'
         }
       },
-      info: {},
-      status: 'error'
+      message: 'Bad Request',
+      url: '/_health/probes/readiness',
+      at: expect.any(String)
```
fuxingloh, line 65: this health.e2e.ts should not be overridden because it provides health check capability for triage.

src/module.api/constants/exception.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const FORBIDDEN_ERROR: ErrorMessage = {
+  statusCode: HttpStatus.FORBIDDEN,
+  message: 'The server understood the request, but is refusing to fulfill it.'
+}
+
+export const VALIDATION_ERROR: ErrorMessage = {
+  statusCode: HttpStatus.BAD_REQUEST,
+  message: 'Request body contains error.'
+}
```
fuxingloh, line 37: I really hate the idea and use of constants.

It creates unnecessary code referencing. To see the causality of an error, you need to look up 3 files. The file that threw the error, the error class definition. The constant file. While it could be simply much easier to embed the constant logic into the error class itself and make it part of the logic.

for e.g.
```ts
abstract class ApiError extends HttpException {
// so that you can use polymorphic types
} 

class BadRequestError extends ApiError {
  static DEFAULT_MESSAGE = '...'

  constructor (message?: string) {
    super(message ?? BadRequestError.DEFAULT_MESSAGE, HttpStatus.BadRequest)
  }
}

class ValidationError extends ApiError {
  static DEFAULT_MESSAGE = '...'

  constructor (message?: string, alternativeParams?: example) {
    super(message ?? ValidationError.DEFAULT_MESSAGE, HttpStatus.BadRequest)
  }
}
```

fuxingloh: The idea IMO, it's always better to use language feature like `class` than to use `constants`.

__e2e__/call.e2e.ts
```diff
@@ -26,7 +26,9 @@ it('should 404 with invalid network', async () => {
   expect(res.json()).toEqual({
     error: 'Not Found',
     message: 'Network not found',
-    statusCode: 404
+    statusCode: 404,
+    url: '/v1/mainnet/call/getblockchaininfo',
+    at: expect.any(String)
```
fuxingloh, line 31: I think a better structure would be:

```ts
error: {}, // for error
data: {}, // for return data
page: {} // for getting next return data
```

This sets the semantic straight. Then for error you can do as such:

```ts
error: {
  code: 404,
  error: 'Not Found',
  message: '...',
  url: '...',
  at: '...'
}
```

src/module.api/interfaces/index.ts
```diff
@@ -0,0 +1,7 @@
+export type AppErrorMessage = string | ErrorMessage
+
+export interface ErrorMessage {
+  statusCode: number
+  message: string
+  error?: any
+}
```
fuxingloh, line 7: As per the constants conversation, you can have the logic wholesomely located together. You don't need a separate interface to pass error. The error details should be the error (Error class) itself.

> FYI: this is another code reference point, the cleaner your code is the less it spiderweb into files. It should be self contained.

src/module.api/constants/exception.constant.ts
fuxingloh: You could also drop the `HTTP` from the error semantic. This keeps the error handling agnostic of the messaging gateway.

src/module.api/filters/exception.filter.ts
```diff
@@ -0,0 +1,27 @@
+import { ExceptionFilter, Catch, ArgumentsHost, HttpException, HttpStatus } from '@nestjs/common'
+import { AppErrorMessage } from '../interfaces'
+
+@Catch(HttpException)
```
fuxingloh, line 4: Should set up 3 sets of filters for a more consistent error handling contract.

1. The error specs you going to use:  e.g. `@Catch(ApiError)` for all implementations of `abstract ApiError`
2. Generic `@Catch(HttpException)`, to parse and wrap into ApiError and send out.
3. Catch all filter, to parse all unknown error types into e.g. ApiEror but state '500', with a simple message of just 'unknown error' so we don't leak sensitive data.

fuxingloh: Hence
1. Known Error with known and defined code/message
2. Compatible Error with guessable/known code/message
3. Unknown Error with unknown message, hence code 500

fuxingloh (changes requested):

src/module.api/constants/exception.constant.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const NOT_FOUND_ERROR: ErrorMessage = {
+  statusCode: HttpStatus.NOT_FOUND,
+  message: 'Resource not found.'
+}
+
+export const BAD_REQUEST_ERROR: ErrorMessage = {
+  statusCode: HttpStatus.BAD_REQUEST,
+  message: 'The request could not be understood by the server due to malformed request'
```
jingyi2811, line 26: Missing fullstop

PR #58, opened 2021-05-10
error handling implementation

### LINKED ISSUE #22: clean error handling paradigm/semantics

fuxingloh:

/kind feature

A clean exception/error handling paradigm. 

At its root, a response body look like this:
```ts
export interface ApiResponse {
  data?: any
  page?: ApiPage
  error?: ApiError
}
```

any mention of `error` indicates an exception occurred.

```ts
export interface ApiError extends Record<string, any> {
  code: HttpStatus
  type: string
  at: number

  /**
   * Optional message to send out.
   */
  message?: string

  /**
   * Optional injected url that produced ApiError.
   * Injected by exception.interceptor.ts.
   */
  url?: string
}
```

Special errors can have their own property inside ApiException for rich features.

```ts
export class ValidationApiException extends ApiException {
  constructor (errors: ValidationError[]) {
    super({
      code: HttpStatus.UNPROCESSABLE_ENTITY,
      type: ApiErrorType.ValidationError,
      at: Date.now(),
      validation: {
        properties: errors.map(ValidationApiException.mapProperty)
      }
    })
  }

  static mapProperty (error: ValidationError): ApiValidationProperty {
    ...
  }
}
```

error mappings are automatically picked up by `exception.interceptor.ts`.

```ts
export class ExceptionInterceptor implements NestInterceptor {
  map (err: Error): ApiException {
    if (err instanceof ApiException) {
      return err
    }

    if (err instanceof HttpException) {
      return new NestJSApiException(err)
    }

    if (err instanceof JellyfishApiError) {
      return new BadRequestApiException(err.message)
    }

    return new UnknownApiException()
  }
}
```

This allows the controller implementor to throw `ApiException` for rich error, `HttpException` for it to automatically remap into `ApiException`, special errors such as `JellyfishApiError` can be remapped. Lastly unknown errors are ignored, remapped as `UnknownApiException`.

Fixes #22


__e2e__/call.e2e.ts
```diff
@@ -40,11 +42,13 @@ it('should 403 with non whitelisted method', async () => {
 
   expect(res.statusCode).toBe(403)
   expect(res.json()).toEqual({
-    error: 'Forbidden',
-    message: 'RPC method not whitelisted',
-    code: 403,
-    url: '/v1/regtest/call/getbalance',
-    at: expect.any(String)
+    error: {
+      code: 403,
```
jingyi2811, line 46: The object sequence of error object should be in the same order as other exception

src/module.api/interceptors/api.error.ts
```diff
@@ -0,0 +1,160 @@
...
+  url?: string
+}
+
+export class ApiException extends HttpException {
+  protected constructor (spec: ApiError) {
+    super({ error: spec }, spec.code)
+  }
+
+  /**
+   * @param {string} url to inject into ApiException
```
jingyi2811, line 45: Fullstop

src/module.api/interceptors/api.error.ts
```diff
@@ -0,0 +1,160 @@
...
+    super({
+      code: HttpStatus.INTERNAL_SERVER_ERROR,
+      type: ApiErrorType.UnknownError,
+      at: Date.now()
+    })
+  }
+}
+
+/**
+ * Intercept NestJS exception and map into ApiException
```
jingyi2811, line 97: Fullstop

src/module.api/interceptors/api.slice.response.ts
```diff
@@ -1,11 +1,19 @@
-import { Response } from '@src/module.api/interceptors/response.interceptor'
+import { ApiResponse } from '@src/module.api/interceptors/response.interceptor'
 
 /**
- * SliceResponse, implements Response interface from response.interceptor.
+ * ApiPage for pagination ApiSliceResponse pagination
```
jingyi2811, line 4: Possible full stop or comma

src/module.api/interceptors/api.slice.response.ts
```diff
@@ -74,16 +80,16 @@ export class SliceResponse<T> implements Response {
...
-  static next<T> (data: T[], next?: string): SliceResponse<T> {
-    return new SliceResponse<T>(data, next)
+  static next<T> (data: T[], next?: string): ApiSliceResponse<T> {
+    return new ApiSliceResponse<T>(data, next)
   }
 
   /**
    * @param {T[]} data array slice
    * @param {number} limit number of elements in the data array slice
    * @param {(item: T) => string} nextProvider to get next token when (limit === data array slice)
```
jingyi2811, line 90: Need to add return

src/module.api/interceptors/api.error.ts
```diff
@@ -0,0 +1,160 @@
...
+    })
+  }
+}
+
+export class NotFoundApiException extends ApiException {
+  constructor (message?: string) {
+    super({
+      code: HttpStatus.NOT_FOUND,
+      type: ApiErrorType.NotFound,
+      at: Date.now(),
```
jingyi2811, line 69: Consistent on object ordering -> code, type, message, at

fuxingloh: I will follow the interface structure instead.

__e2e__/call.e2e.ts
fuxingloh: don't need to be that strict on tests, it tends to halt progress. For code it's okay.

src/module.api/interceptors/api.error.ts
fuxingloh: The usage of punctuations in the English language aids the reader understanding of the body of text and the information conveyed. Often used to separate written sentences into parts of sentences. And to that effect, they are to separate sentences. In writing; a listicle conveys information that is already separated into parts. `@param` is a form of a listicle.

Not all written work requires a 'period' as they do not carry much information that necessitated the need for the use of punctuation. Documentation in code has similar standards, as they don't carry enough information to form a full sentence. 

`{string} URL to inject into ApiException`
vs
`A string URL is injected into ApiException that will be intercepted by ApiExceptionInterceptor and added into the response body.`

One is a proper sentence and one isn't, this theme of writing is very common in programming documentation as it does not require a full sentence.

PR #59, opened 2021-05-11
added packages/whale-api-client

fuxingloh:
/kind feature

This moves the `whale-api-client` (therefore also `whale-api-wallet`) from jellyfish to whale. As the development of the DeFi whale project is closely iterative together with DeFi Whale. It only makes sense to have them together and to be released together. This allows us to be the consumer of our own client implementation. (Dogfooding at its finest.) 

The benefit of such implementation practices enforced at CI allows us to end-to-end test our services. Hence we can eliminate `__e2e__` to just test on `packages/whale-api-client`. The testing requirement should be the same as e2e.

With the CD integrated, when we release a new DeFi Whale release. A client and docker image will be produced aligned to the same version of jellyfish.

I skimmed on exhaustive testing as necessitated by my ability to move fast. If you happen to chance upon the code while adding features please do help me add relevant documentation and more testing.

Fixes https://github.com/DeFiCh/jellyfish/issues/138

#### Additional comments?:
- lerna created 27k `package-lock.json` changes, WOW


packages/whale-api-client/__tests__/api/call.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  await service.start()
+})
+
+afterAll(async () => {
+  await service.stop()
+  await container.stop()
+})
+
+it("call('getblockchaininfo')", async () => {
+  const info = await client.call.call<blockchain.BlockchainInfo>('getblockchaininfo', [], 'number')
```
jingyi2811, line 27: call.call should be call

fuxingloh: currently this is as expected to change this we will need to refactor the endpoint name, I will do that in the future.

canonbrother (approved):

jingyi2811 (approved):

Issue #82, opened 2021-05-18
prefix on cache keys


canonbrother:
#### What would you like to be added:
Enumeration to construct a prefix on cache key

#### Why is this needed:
Safety wise, its better to prefix the cache key `eg: token_${key}, poolpair_${key}` which provides a scope protection that helps preventing key overwritten

fuxingloh:
/assign @fuxingloh 
/priority urgent-now

I will work on this now, should be a quick one

fuxingloh:
Wrote the specification without testing in https://github.com/DeFiCh/whale/tree/fuxingloh/cache

/assign @canonbrother can you help take over this, I am a little too occupied now.

PR #83, opened 2021-05-18
added poolpairs.controller.ts

### LINKED ISSUE #6: poolpair.controller.ts

canonbrother:

/kind

Fixes #6

### DIFF
--- a/packages/whale-api-client/__tests__/api/poolpair.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpair.test.ts
@@ -135,7 +135,7 @@ describe('get', () => {
     const response = await client.poolpair.get('7')
 
     expect(response).toEqual({
-      id: '7',
+      id: 7,
       symbol: 'A-B',
       name: 'A-B',
       status: true,
@@ -154,7 +154,6 @@ describe('get', () => {
       tradeEnabled: true,
       ownerAddress: expect.any(String),
       rewardPct: 0,
-      customRewards: undefined,
       creation: {
         tx: expect.any(String),
         height: expect.any(Number)
--- a/packages/whale-api-client/__tests__/api/poolpair.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpair.test.ts
@@ -3,7 +3,6 @@ import { StubWhaleApiClient } from '../stub.client'
 import { StubService } from '../stub.service'
 import { WhaleApiClient } from '../../src'
 import { createPoolPair, createToken, addPoolLiquidity, getNewAddress, mintTokens } from '@defichain/testing'
-import BigNumber from 'bignumber.js'
 
 let container: MasterNodeRegTestContainer
 let service: StubService
@@ -142,23 +141,23 @@ describe('get', () => {
       status: true,
       tokenA: {
         id: expect.any(String),
-        reserve: new BigNumber('100'),
-        blockCommission: new BigNumber('0')
+        reserve: 100,
+        blockCommission: 0
       },
       tokenB: {
         id: expect.any(String),
-        reserve: new BigNumber('200'),
-        blockCommission: new BigNumber('0')
+        reserve: 200,
+        blockCommission: 0
       },
-      commission: new BigNumber('0'),
-      totalLiquidity: new BigNumber('141.42135623'),
+      commission: 0,
+      totalLiquidity: 141.42135623,
       tradeEnabled: true,
       ownerAddress: expect.any(String),
-      rewardPct: new BigNumber('0'),
+      rewardPct: 0,
       customRewards: undefined,
       creation: {
         tx: expect.any(String),
-        height: expect.any(BigNumber)
+        height: expect.any(Number)
       }
     })
   })
--- a/packages/whale-api-client/__tests__/api/poolpair.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpair.test.ts
@@ -79,22 +79,20 @@ describe('list', () => {
       symbol: 'A-C',
       name: 'A-C',
       status: true,
-      token: {
-        a: '1',
-        b: '3'
+      tokenA: {
+        id: '1',
+        reserve: 50,
+        blockCommission: 0
       },
-      reserve: {
-        a: 50,
-        b: 300
+      tokenB: {
+        id: '3',
+        reserve: 300,
+        blockCommission: 0
       },
       commission: 0,
       totalLiquidity: 122.47448713,
       tradeEnabled: true,
       ownerAddress: expect.any(String),
-      blockCommission: {
-        a: 0,
-        b: 0
-      },
       rewardPct: 0,
       customRewards: undefined,
       creation: {
@@ -141,22 +139,20 @@ describe('get', () => {
       symbol: 'A-B',
       name: 'A-B',
       status: true,
-      token: {
-        a: expect.any(String),
-        b: expect.any(String)
+      tokenA: {
+        id: expect.any(String),
+        reserve: new BigNumber('100'),
+        blockCommission: new BigNumber('0')
       },
-      reserve: {
-        a: new BigNumber('100'),
-        b: new BigNumber('200')
+      tokenB: {
+        id: expect.any(String),
+        reserve: new BigNumber('200'),
+        blockCommission: new BigNumber('0')
       },
       commission: new BigNumber('0'),
       totalLiquidity: new BigNumber('141.42135623'),
       tradeEnabled: true,
       ownerAddress: expect.any(String),
-      blockCommission: {
-        a: new BigNumber('0'),
-        b: new BigNumber('0')
-      },
       rewardPct: new BigNumber('0'),
       customRewards: undefined,
       creation: {
--- a/packages/whale-api-client/__tests__/api/poolpair.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpair.test.ts
@@ -158,35 +158,3 @@ describe('get', () => {
     await expect(client.poolpair.get('B-Z')).rejects.toThrow()
   })
 })
-
-describe('listPoolShares', () => {
-  it('should listPoolShares', async () => {
-    const response = await client.poolpair.listPoolShares(30)
-
-    expect(response.length).toBe(3)
-    expect(response.hasNext).toBe(false)
-
-    expect(response[1]).toEqual({
-      poolID: '8',
-      owner: expect.any(String),
-      percent: 99,
-      amount: 122.47447713,
-      totalLiquidity: 122.47448713
-    })
-  })
-
-  it('should listPoolShares with pagination', async () => {
-    const first = await client.poolpair.listPoolShares(2)
-    expect(first.length).toBe(2)
-    expect(first.hasNext).toBe(true)
-    expect(first.nextToken).toBe('8')
-
-    expect(first[0].poolID).toBe('7')
-    expect(first[1].poolID).toBe('8')
-
-    const next = await client.paginate(first)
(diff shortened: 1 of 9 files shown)


packages/whale-api-client/__tests__/api/poolpair.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    expect(last.nextToken).toBeUndefined()
+
+    expect(last[0].symbol).toBe('B-D')
+    expect(last[1].symbol).toBe('B-E')
+  })
+})
+
+describe('get', () => {
+  it('should get', async () => {
+    const response = await client.poolpair.get('A-B')
```
canonbrother, line 132: not sure why here getting `Pool not found`

src/module.api/poolpair.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+   * @param {string} symbol
+   * @return {Promise<PoolPairInfo>}
+   */
+  @Get('/:symbol')
+  async get (symbol: string): Promise<PoolPairInfo> {
+    const poolPairInfo = await this.poolPairInfoCache.get(symbol)
+    if (poolPairInfo === undefined) {
+      throw new ConflictException('unable to find token')
+    }
+    return poolPairInfo
```
fuxingloh, line 52: We need to use `id: string` instead of `symbol` because when the operator feature kicks in there will be multiple competing PoolPair by different operators.

fuxingloh: Oh and don't use ConflictException for NotFound error.

fuxingloh: Conflict is should have but NotFound, NotFound is just NotFound.

src/module.api/poolpair.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+    }, true)
+
+    const poolShareInfosDto = Object.entries(poolShareResult).map(([, value]) => {
+      return mapPoolShare(value)
+    }).sort(a => Number.parseInt(a.poolID))
+
+    return ApiPagedResponse.of(poolShareInfosDto, query.size, item => {
+      return item.poolID
+    })
+  }
```
fuxingloh, line 78: Best to create another PR for this, separate this out because that first 2 endpoint is more urgent. We might want to have another PR to discuss how this should be implemented.

src/module.api/poolpair.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+    'reserveA/reserveB': poolPairInfo['reserveA/reserveB'],
+    'reserveB/reserveA': poolPairInfo['reserveB/reserveA'],
+    tradeEnabled: poolPairInfo.tradeEnabled,
+    ownerAddress: poolPairInfo.ownerAddress,
+    blockCommissionA: poolPairInfo.blockCommissionA,
+    blockCommissionB: poolPairInfo.blockCommissionB,
+    rewardPct: poolPairInfo.rewardPct,
+    customRewards: poolPairInfo.customRewards,
+    creationTx: poolPairInfo.creationTx,
+    creationHeight: poolPairInfo.creationHeight
```
fuxingloh, line 102: I was thinking whether we can remap to something like this:

```ts
tokenA: {
  id: ,
  reserve: ,
  commission: ,
},
tokenB: {
...
}
```

And if `reserveA/reserveB` can be computed ownself, we can remove it.

src/module.api/cache/poolpair.info.cache.ts
```diff
@@ -0,0 +1,40 @@
...
+    let poolPairInfo = await this.cacheManager.get<PoolPairInfo>(symbol)
+
+    if (poolPairInfo !== undefined) {
+      return poolPairInfo
+    }
+
+    const poolPairResult = await this.rpcClient.poolpair.getPoolPair(symbol, true)
+    for (const k in poolPairResult) {
+      poolPairInfo = poolPairResult[k]
+    }
```
fuxingloh, line 32: should separate this into another method, easier to read

packages/whale-api-client/__tests__/api/poolpair.test.ts
fuxingloh: Can't see why also, try running in debug mode?

src/module.api/cache/poolpair.info.cache.ts
canonbrother: hmm.. perhaps i make it one line
```
poolPairInfo = poolPairResult[ Object.keys(poolPairResult)[0] ]
```

src/module.api/poolpair.controller.ts
canonbrother: i actually follow this https://github.com/DeFiCh/whale/blob/8af244c50118436567ce340f316f0bdb7b49efeb/src/module.api/address.controller.ts#L40
oh yea i just copy paste it.. the message was wrong as well

yea `Conflict` is normally used for crashing with existing item right??

fuxingloh: For that use case because the token exists in your account, but cannot be found in the cache. Hence a conflict.

src/module.api/cache/poolpair.info.cache.ts
fuxingloh: I meant this:

https://github.com/DeFiCh/whale/blob/8af244c50118436567ce340f316f0bdb7b49efeb/src/module.api/cache/token.info.cache.ts#L27-L54

Because if you put everything together all the concerns get tied up into one function, harder to maintain in the long run. 

All mixed up, for one logical branch:
1. get: cache read, cache miss, **live read**, cache put, return put
1. get: cache read, cache hit, return hit

Live read is a heavy code branch that is part of another code logical branch that should be separated out. 

> https://en.wikipedia.org/wiki/Category_theory Categories represent abstractions of other mathematical concepts. Many areas of mathematics can be formalised by category theory as categories. Hence **category theory uses abstraction to make it possible to state and prove many intricate and subtle mathematical results in these fields** in a much simpler way.

packages/whale-api-client/__tests__/api/poolpair.test.ts
canonbrother: use `get by id`.. solved.. =.=""

src/module.api/poolpair.controller.ts
```diff
@@ -44,13 +44,16 @@ export class PoolPairController {
    * @return {Promise<PoolPairInfoDto>}
    */
   @Get('/:id')
-  async get (id: string): Promise<PoolPairInfoDto> {
+  async get (@Param('id') id: string): Promise<PoolPairInfoDto> {
```
canonbrother, line 47: bug: missing `@Param`    =.=""

fuxingloh: Wow, hahaha

src/module.api/poolpair.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const poolPairResult = await this.rpcClient.poolpair.getPoolPair(id)
+    if (poolPairResult === undefined) {
+      throw new NotFoundException('unable to find poolpair')
+    }
+    poolPairInfoDto = mapPoolPair(id, poolPairResult[id])
+
+    await this.poolPairInfoCache.set(id, poolPairInfoDto)
+
+    return poolPairInfoDto
```
fuxingloh, line 61: I think there is still a better way to tackle this but it's beyond this scope. I guess we will look into it when we do https://github.com/DeFiCh/whale/issues/82

src/module.api/poolpair.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+      return item.id
+    })
+  }
+
+  /**
+   * @param {string} id
+   * @return {Promise<PoolPairInfoDto>}
+   */
+  @Get('/:id')
+  async get (@Param('id') id: string): Promise<PoolPairInfoDto> {
```
fuxingloh, line 47: You could add a ParsetIntPipe for Param to validate path

src/module.api/poolpair.controller.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      customRewards: undefined,
+      creation: {
+        tx: expect.any(String),
+        height: expect.any(BigNumber)
+      }
+    })
+  })
+
+  it('should throw error while getting non-existent poolpair', async () => {
+    await expect(controller.get('999')).rejects.toThrow()
```
fuxingloh, line 181: Should check the error type `NotFoundException`

packages/whale-api-client/src/api/poolpair.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+export interface PoolShareInfoDto {
+  poolID: string
+  owner: string
+  percent: BigNumber
+  amount: BigNumber
+  totalLiquidity: BigNumber
+}
```
fuxingloh, line 67: should remove this first.

packages/whale-api-client/src/api/poolpair.ts
```diff
@@ -0,0 +1,67 @@
...
+   *
+   * @param {string} id
+   * @return {Promise<PoolPairInfoDto>}
+   */
+  async get (id: string): Promise<PoolPairInfoDto> {
+    return await this.client.requestData('GET', `poolpairs/${id}`)
+  }
+}
+
+export interface PoolPairInfoDto {
```
fuxingloh, line 34: For `packages/*` best to just name it `PoolPairData`.

1. packages are for public consumption hence, Dto is not a commonly used term outside of NestJS, only applicable within NestJS
2. `Info` is kind of fact conveying terminology, not good for conveying data states
3. We can't keep it as `PoolPair` or else it might conflict with the `export class PoolPair {`
3.  I use `PoolPairData` as data is a good word to represent the data and its state.
3. Data is also commonly understandable

packages/whale-api-client/__tests__/api/poolpair.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      customRewards: undefined,
+      creation: {
+        tx: expect.any(String),
+        height: expect.any(Number)
+      }
+    })
+  })
+
+  it('should throw error while getting non-existent poolpair', async () => {
+    await expect(client.poolpair.get('999')).rejects.toThrow()
```
fuxingloh, line 166: should test error message

fuxingloh (changes requested):

src/module.api/poolpair.controller.ts
canonbrother: hmm... it would be considered call `fat` controller

fuxingloh (approved):

PR #135, opened 2021-06-02
error type handling, added more tests and refactoring

### LINKED ISSUE #91: missing `NotFoundException` testing for `address.getBalance` and `address.getAggregation`

Test `NotFoundException` condition, missing in https://github.com/DeFiCh/whale/pull/49#issuecomment-844749831

#### Why is this needed:

For completeness sake.

### LINKED ISSUE #99: `NotFoundException` must be for NotFound errors
#### What happened:

`NotFoundException` is used for `catch(err)` which is a catch-all error. Need to parse and verify it's actually a not found error.

canonbrother:

/kind feature

1. specify error with right exception + output meaningful message + update test
2. added some missing tests
3. some refactors
  - better error test pattern, refactor some `toThrow` pattern to correct `try.. catch` pattern, latter is more informative
  - file naming (singular.file.ts) follows angular style guide as nestjs is inspired by angular [style guide](https://angular.io/guide/styleguide#general-naming-guidelines)
  - some funcs naming, eg: list{singular} -> list{plural}
  - test grouping `describe`, hmm.. currently i standardize most of the tests rely on 1 container first, will optimize it

Fixes #91
Fixes #99


src/module.api/shared/utils.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * To grab object value by key
+ *
+ * @param {any} obj
+ * @param {string} key
+ * @return {any}
+ */
+export function grabValueByKey (obj: any, key: string): any {
```
fuxingloh, line 8: We can remove shared/utils with `lodash`.

1. https://lodash.com/docs/4.17.15#get
2. https://lodash.com/docs/4.17.15#isObject

src/module.api/transaction.controller.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+          code: 400,
+          type: 'BadRequest',
+          message: 'Transaction decode failed',
+          at: expect.any(Number)
+        })
+      }
+    })
+
+    it('should throw BadRequestError due to high fees', async () => {
+      const hex = await createSignedTxnHex(container, 10, 9)
```
fuxingloh, line 77: missing `throw` or  `expect.assertions(2)` here

fuxingloh (changes requested):
thanks for the refactoring!!! 

Some comments:

1. We can use `lodash` for utilities so we don't need to maintain them. 
2. There are some `throw error` tests with missing `throw new Error('do not reach here')` but I was thinking maybe adding `expect.assertions(2)` would be cleaner

src/module.api/shared/utils.ts
fuxingloh: May packages already import them, we can import them also.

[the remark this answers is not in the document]
canonbrother: removed

canonbrother: ohh.. just noted that the `lodash` is a rich js utilities lib.

previously the purpose i created the `grabValueByKey` was to easily get the value by vary position of key (nested)
grabValueByKey(e, 'message')
```
e.message
e.payload.message
```
hmm... i think i'm a bit over-engineered, so straight away get the value will do.

packages/whale-api-client/__tests__/api/poolpair.test.ts
```diff
@@ -162,12 +162,36 @@ describe('get', () => {
   })
 
   it('should throw error as numeric string is expected', async () => {
-    await expect(client.poolpair.get('A-B')).rejects.toThrow(WhaleApiException)
-    await expect(client.poolpair.get('A-B')).rejects.toThrow('Validation failed (numeric string is expected)')
+    expect.assertions(2)
+    try {
+      await client.poolpair.get('A-B')
+      throw new Error('should not reach here')
+    } catch (err) {
```
fuxingloh, line 169:
```suggestion
    expect.assertions(2)
    try {
      await client.poolpair.get('A-B')
    } catch (err) {
```

With `expect.assertions` you don't need `throw new Error` since it does the same thing. I think using expect.assertions is clean when doing try catch test assertions. Since `throw new Error` is kind of anomalous.

fuxingloh (changes requested):

fuxingloh (approved):

PR #137, opened 2021-06-07
added prefix for cache

### LINKED ISSUE #82: prefix on cache keys

canonbrother:

/kind feature
/kind refactor

To categorize the cache for easier to differentiate cache id by prefix and prevent get overwritten by others

Fixes #82


fuxingloh:
@canonbrother take a look at this, https://github.com/DeFiCh/whale/commit/87e7b46f0019ca874e435d6445641621ab9dd0b7

canonbrother:
> @canonbrother take a look at this, [87e7b46](https://github.com/DeFiCh/whale/commit/87e7b46f0019ca874e435d6445641621ab9dd0b7)
> 
> I think we should do something like this.

I see... yea noted!!

src/module.api/cache/global.cache.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * For testing purpose
+   *
+   * @param key of cache
+   * @return {Promise<T | undefined>}
+   */
+  async testGet<T> (key: string): Promise<T | undefined> {
+    return await this.cacheManager.get<T>(key)
```
fuxingloh, line 60: Since there is only one global cache, you can get the same instance via `module.get<Cache>(Cache)`.

src/module.api/cache/defid.cache.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    // TODO(canonbrother): add listpoolpairs in @defi/testing
+    const poolPairResult = await container.call('listpoolpairs')
+    for (const k in poolPairResult) {
+      await cache.getPoolPairInfo(k)
+    }
+
+    await container.stop()
+  })
+
```
fuxingloh, line 48: I am curious about cache miss behavior I think we need coverage for that. Codecov is reporting that we don't have test for that scenario also.

src/module.api/cache/defid.cache.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+let cache: DeFiDCache
+
+describe('getPoolPairInfo', () => {
+  beforeAll(async () => {
+    await container.start()
+    await container.waitForReady()
+    await container.waitForWalletCoinbaseMaturity()
+    client = new JsonRpcClient(await container.getCachedRpcUrl())
+
+    const module = await Test.createTestingModule({
```
fuxingloh, line 21: make this global

src/module.api/cache/defid.cache.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    for (const k in poolPairResult) {
+      await cache.getPoolPairInfo(k)
+    }
+
+    await container.stop()
+  })
+
+  it('should get from cache via get as container RPC is killed', async () => {
+    const abKey = `${CachePrefix.POOL_PAIR_INFO} 4`
+    const abPoolPair = await cache.testGet<PoolPairInfo>(abKey)
```
fuxingloh, line 51:
```suggestion
    const cache = module.get<Cache>(Cache)
    const abPoolPair = await cache.get<PoolPairInfo>(abKey)
```

Should be able to return to do this.

fuxingloh (changes requested):

src/module.api/cache/defid.cache.spec.ts
canonbrother: the result after killed the rpc and cache reset
```
try {
      await defiCache.getPoolPairInfo('4')
} catch (err) {
      expect(err.code).toStrictEqual('ECONNREFUSED')
      expect(err.type).toStrictEqual('system')
}
```

canonbrother: hmm... wondering how to test this sanity check? 
it will get undefined only while blockchain output undefined
properly need to mock and hard return undefined
currently i skip it first
```
const result = await this.rpcClient.poolpair.getPoolPair(id)
/* istanbul ignore if  */
if (result[id] === undefined) {
    return undefined
}
```

src/module.api/cache/defid.cache.ts
```diff
@@ -0,0 +1,57 @@
...
+    return await this.get<PoolPairInfo>(CachePrefix.POOL_PAIR_INFO, id, this.fetchPoolPairInfo.bind(this))
+  }
+
+  private async fetchPoolPairInfo (id: string): Promise<PoolPairInfo | undefined> {
+    const result = await this.rpcClient.poolpair.getPoolPair(id)
+
+    /* istanbul ignore if  */
+    if (result[id] === undefined) return undefined
+
+    return result[id]
```
fuxingloh, line 55: I think there are too many if conditions, creating too many logical branches.

We should make fetchPoolPairInfo return `PoolPairInfo | undefined`, for errors like "Unable to find poolpair" we can catch it here and return undefined. We can ignore other errors and escalate. This will simplify the consumer side, making it easier for them to consume the cache without worrying about error handling.

@canonbrother what do you think? I think this will cut down how deep our vertical codes get

```diff
-    try {
      const info = await this.deFiDCache.getPoolPairInfo(id)
      if (info === undefined) {
+        throw new NotFoundException('Unable to find poolpair')
      }
      return mapPoolPair(String(id), info)
-    } catch (e) {
-      /* istanbul ignore else */
-      if (e?.payload?.message === 'Pool not found') {
-        throw new NotFoundException('Unable to find poolpair')
-      } else {
-        throw new BadRequestException(e)
-      }
-    }
```

fuxingloh (changes requested):
some suggestions for review

src/module.api/cache/defid.cache.ts
canonbrother: hmm.. 
```
const info = await this.deFiDCache.getPoolPairInfo(id)
if (info === undefined) { // <-- this is not reachable as rpc will throw error beforehand
   throw new NotFoundException('Unable to find poolpair')
}
 ```

Comment (fuxingloh) [reply src/module.api/cache/defid.cache.ts:55]: My thoughts are we keep the "expected returns" (errors included) within the cache. Meaning we catch the errors within the cache so that we return undefined if "Unable to find poolpair" are raised from defid-cache. As not found is an expected error, this makes it easier to consume the defid-cache without having to handle all the expected/unexpected errors.

Comment (canonbrother) [reply src/module.api/cache/defid.cache.ts]: yea done.. i refactored it by following
1. [deficache] try.. catch on exactly rpc call.. if err return undefined
2. [controller] if undefined throw err

File: src/module.api/cache/defid.cache.ts [hunk 6]
```diff
@@ -0,0 +1,60 @@
...
+
+  async getPoolPairInfo (id: string): Promise<PoolPairInfo | undefined> {
+    return await this.get<PoolPairInfo>(CachePrefix.POOL_PAIR_INFO, id, this.fetchPoolPairInfo.bind(this))
+  }
+
+  private async fetchPoolPairInfo (id: string): Promise<PoolPairInfo | undefined> {
+    try {
+      const result = await this.rpcClient.poolpair.getPoolPair(id)
+      return result[id]
+    } catch (err) {
```
fuxingloh, line 55: you still need to escalate non NotFound exception in defid.cache.ts

fuxingloh (changes requested):

src/module.api/cache/defid.cache.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (result[id] === undefined) {
+        return undefined
+      }
+      return result[id]
+    } catch (err) {
+      if (err?.payload?.message === 'Pool not found') {
+        throw new NotFoundException('Unable to find poolpair')
+      } else {
+        throw new BadRequestException(err)
+      }
```
fuxingloh, line 68:
```suggestion
      if (err?.payload?.message === 'Pool not found') {
        return undefined
      } else {
        throw err
      }
```

src/module.api/cache/defid.cache.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (tokens[0] === undefined) {
+        return undefined
+      }
+      return tokens[0]
+    } catch (err) {
+      if (err?.payload?.message === 'Token not found') {
+        throw new NotFoundException('Unable to find token')
+      } else {
+        throw new BadRequestException(err)
+      }
```
fuxingloh, line 48:
```suggestion
      if (err?.payload?.message === 'Token not found') {
        return undefined
      } else {
        throw err
      }
```

fuxingloh (changes requested):

src/module.api/cache/defid.cache.ts
canonbrother: phew.. got it!!

fuxingloh (approved):

PR #149, opened 2021-06-14
Add OracleController

### LINKED ISSUE #144: `Oracles` Service/Controller

/triage accepted
/assign @jingyi2811 
/priority important-soon

/area module-api

jingyi2811:

/kind feature


jingyi2811:
This is not yet completed.

For controller,

I have added 1 test function for all 9 oracle rpcs.
I will add 2 more test functions, which cover 
- UTXO 
- fail scenarios

I took a longer time to finish the 1st test function. But the same code with some changes will be used to create the other 2 functions

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @Put('remove/:id')
+  async removeOracle (@Param('id', ParseIntPipe) id: string, @Body() utxos?: UTXO[]): Promise<any> {
+    return await this.client.oracle.removeOracle(id, utxos)
+  }
+
+  @Put('update/:id')
+  async updateOracle (@Param('id', ParseIntPipe) id: string, address: string, @Body() updateOracleOptions: UpdateOracleOptions): Promise<any> {
+    return await this.client.oracle.updateOracle(id, address, updateOracleOptions)
+  }
```
fuxingloh, line 29: remove all mutating RPC, OracleController must only provide stateless service

fuxingloh (changes requested):

jingyi2811:
I write the code at a fast pace. Added mock pagination functions to the 3 rpcs below in order to let my test passed. Won't take a closer look at every line that I wrote until the paginations feature is added to the C++ code.

- listoracles
- listlatestrawprices 
- listprices

jingyi2811:
[It is a bit late now, I will analyze and write more tomorrow]

On the blockchain, users can only query the oracle prices for timestamps 1 hour before the current time.
For all the historical prices created any time before, users need to query the database instead.

4 database tables will be created to store ALL the prices (including the historical prices + future prices [which is happened 1 hour later after the current time]).

- Oracle
- OracleInfo (Basic info)
- OraclePriceFeed  (Token and currency info)
- OraclePrice (Price info)

Oracle
- OracleId (string [primaryKey])
- Status (Enum: active / inactive)

OracleInfo
- OracleInfoId (number [primaryKey auto_increment])
- OracleId (string [foreignKey])
- Address (string)
- Weightage (number)
- Status (Enum: active / inactive)

OraclePriceFeed
- OraclePriceFeedId (number [primaryKey auto_increment])
- OracleId (string [foreignKey])
- Token (string)
- Currency (string)
- Status (Enum: active / inactive)

OraclePrice
- OraclePriceId (number [primaryKey auto_increment])
- OraclePriceFeedId (number [foreignKey])
- Price (number [Allow null])
- Timestamp (number)
- Status (Enum: active / inactive)

Oracle and OracleInfo have  1-to-many relationship.
Oracle and OraclePriceFeed have  1-to-many relationship.
OraclePriceFeed and OraclePrice have 1-to-many relationship.

For all the tables, no record will be deleted forever. 
Interception should happen on 4 RPCS (appointoracle, removeoracle, updateoracle, setoracledata)
When an oracle is removed, the status of the oracle, all its oracleInfos, and oraclePriceFeeds records will be set to inactive.
When an oracle is updated, depending on what is the user input, either OracleInfo or (OraclePriceFeed and OraclePrice) table records will be inserted / updated.

[This is something that needs to be discussed further]

For the OraclePrice table, it is mandatory for ONLY 1 record of a oraclePriceFeedId to be inserted every 5 minutes.  If the timestamp is equivalent to 12:55 am, this means that it is for the period between 12:50.00 am and 12:54:59 am.  if there are 3 prices recorded in the blockchain for that period of time, the system should only save 1 price record (I recommend we save the MEAN of the 3 prices).
If there is no price found in the blockchain, the system should save a NULL value. We should not save 0.

jingyi2811:
[Still not yet done]

On the landing page, all the ACTIVE price feeds are displayed in pagination. 

There are 4 categories of price feed 

- Crypto
- Stock
- Commodity
- Forex

Users can narrow down the price feed search results by selecting any of these 4: category, active/inactive, token, or currency. Users can only select 1 price feed card at a time. 

There are 2 types of views for every price feed: 
- Lite view 
- Detail view

The Lite view displays the current price information only, whereas the detail view displays all the historical price information for the price feed.

Lite view

- Token / Currency
- Current price
- Last 24-hour price percentage difference
- Active / Inactive
- Category

Detail view

- Historical price (24 H, 7 D, 30 D, 90 D, ALL)
- If a point of the chart is selected, the user can see the date time and other Oracle info like

  1.  How many Oracle responses? 
  2. What is the last Oracle updated time?
  3. What is the price of every oracle?

[Extra]

Other than that, the user can update his 

- Preferred currency
- Preferred price list

So the landing page will display his preferred currency or preferred price feed first next time.

jingyi2811:
I am going to start writing the indexer code for 4 database tables. 
If there are some changes required, I can improve the code later.

fuxingloh:
> See
> https://docs.google.com/document/d/1Qat_8ZlDBPwqdKEwPY1g4QofEyIKekxnNQl2HMpROc4/edit?usp=sharing

keep the docs in GitHub PR or GitHub Issue

Things to note:
1. We are not using a traditional database, can't think of it from a traditional RDMS perspective. Look LSM database.
2. No RPCs call, you should be able to index historical data from every RawBlock connected in the indexer
3. You need to use an aggregation technique that rolls forward and creating a moving average to form a historical data set

jingyi2811:
**Product Design**

There are 4 price feed categories

- Crypto (For eg: BTC/USD, ETH/USD)
- Stock (For eg: APPL/USD, TSLA/USD)
- Commodity (For eg: OIL/USD, GOLD/USD)
- Forex (For eg: EUR/USD, SGD/USD)

On the landing page, the system should display all the active price feed cards in pagination. Users can narrow down the selection result by checking any of these 4

- Category
- Status [All, Active (By default), Inactive]
- Token
- Currency

checkboxes or key in the token/currency value directly in the textbox and click the search button. Then, users can only select 1 price feed card at a time. 

Every card has 2 view types

- Lite (Display the current price info)
- Detail (Display all the historical price info)

Lite view
- Token/Currency
- Current price
- Last 24-hour price percentage difference
- Active/Inactive (Inactive = There is no price offered for the past 15 minutes) 
- Category
- Oracle info
  1. How many Oracle responded and contributed to creating the price? What are their names? And when they responded?
  2. What was the last price updated time?
  3. What was the price of every Oracle?

Detail view
- A chart that shows the historical price (24 H, 7 D, 30 D, 90 D, 3M, 6M, 1Y, 5Y, ALL)

Bonus

- Can we allow users to log in and update their preferred currency and price list? So the system can display their preferred data first on the list.
- We probably need to build a few APIs to allow developers to query the historical price.
- We can have some kind of portfolio management, which allows the users to key in and save the token amount they hold, and the system can calculate their total token worth base on current prices.

jingyi2811:
**LSM Database**

We need 2 keys

- PriceFeed {token}-{currency}-{blockheight} : {price}-{timestamp}
- Oracle {oracleid}-{token}-{currency}-{blockheight} : {price}-{weightage}-{timestamp}
 
We only allow GET and PUT operations on both keys. There is no DELETE. If a price feed or oracle was removed from the blockchain, we still keep them in the LSM database.

jingyi2811:
**Indexing**

Interception should happen on setOracleData on every block

We need to get the data similar to 

calling listLatestRawPrice RPC and put In Oracle key
calling listPrice RPC and put in PriceFeed key

jingyi2811:
**Retrieve Data for UI**

On the landing page, the system displays all the price list cards filtered by pagination. `[Get from Price Feed key that exists for the past 1 hour]`

On the landing page, the system should display all the active price feed cards in pagination. Users can narrow down the selection result by checking any of these 4

- Category `For every price feed, we need to hardcode somewhere in the Front End code`
- Status [All, Active (By default), Inactive] `Active means there is price feed record for the past 1 hour`
- Token
- Currency

checkboxes or key in the token/currency value directly in the textbox and click the search button. 

Then, users can only select 1 price feed card at a time. Every card has 2 view types

Lite (Display the current price info)
Detail (Display all the historical price info)

For every card, there are 2 view types: Lite view and Detail view

Lite View

- Token/Currency 
- Current price 
- Last 24-hour price percentage difference 
- Active/Inactive`Active means there is a price feed record for the past 1 hour`
- Category`Hardcode somewhere in the Front End code`
- Oracle info
    1. How many Oracle responded and contributed to creating the price? What are their names? And when they responded?
    2. What was the last price updated time?
    3. What was the price of every Oracle?
`Get the value from Oracle key`


Detail view
A chart that shows the historical price (24 H, 7 D, 30 D, 90 D, 3M, 6M, 1Y, 5Y, ALL)
`Get the value from PriceFeed key`

PR #150, opened 2021-06-15
fix invalidate loop unable to fix raw block

fuxingloh:

/kind fix

When RawBlock isn't indexed before invalidating kicks in, it will get stuck in invalidate loop. This fixes it.


fuxingloh:
going to merge, need to this for testing

PR #151, opened 2021-06-15
poolpair.service extra stats

### LINKED ISSUE #136: `PoolPair` Service/Controller Extra Stats

Extending on top of the current `PoolPairController` via `PoolPairService`, we need to display statistical data similar to how it's done in https://dex.defichain.com/mainnet/pool but in a decentralized implementation that is resilient to network cascade.

- Price Ratio
- APR
- Liquidity (USD)
- Daily Volume (USD)
  - 30 Day Average
  - 1 Day 
- Total Liquidity (Pairs)

We don't need external service for this, (USD) can be calculated by the USDT-DFI pair. We can derive all value from that pair in real-time. As `USDT` pair configuration can vary in different networks and for the purpose of testing, we should be able to set that value in `configuration.ts`. This allows us to use varying base pair for different currency calculation as the base.

This should be implemented with caching, (blocked by #82) and the logic should be separated into PoolPairService for good practices. 

/triage accepted
/assign @canonbrother 
/priority important-soon

Areas of concerns are, but might not be limited to:
/area module-api module-indexer module-database module-model
Note that there is technical debt in the `module-indexer` and `module-model` due to the crunch, they are not the cleanest code.

#### Why is this needed:

For Wallet & Explorer to display statistical data.

## `sub stories`
- [x] add priceRatio and totalLiquidityUsd
- [ ] poolswap block sync, aggregation and index
- [ ] add apy and volume (avg 30 days)

canonbrother:

/kind feature

Fixes #136

### DIFF
--- a/src/module.api/_module.ts
+++ b/src/module.api/_module.ts
@@ -6,6 +6,7 @@ import { TransactionsController } from '@src/module.api/transaction.controller'
 import { ApiValidationPipe } from '@src/module.api/pipes/api.validation.pipe'
 import { AddressController } from '@src/module.api/address.controller'
 import { PoolPairController } from '@src/module.api/poolpair.controller'
+import { PoolPairService } from '@src/module.api/poolpair.service'
 import { TokenInfoCache } from '@src/module.api/cache/token.info.cache'
 import { PoolPairInfoCache } from '@src/module.api/cache/poolpair.info.cache'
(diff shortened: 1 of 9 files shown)


canonbrother:
close
refer to better fix
https://github.com/DeFiCh/whale/pull/202

PR #177, opened 2021-06-23
Add oracle controller

### LINKED ISSUE #144: `Oracles` Service/Controller

jingyi2811:
What kind of PR is this?:
/kind feature

Fixes #144

Additional comments?:


src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,103 @@
...
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (!vout.scriptPubKey.hex.startsWith('6a')) {
+          continue
+        }
+
+        const priceSet = new Set()
+
+        try {
+          const prices = await this.priceMapper.getActive(block.time)
```
jingyi2811, line 29: In the final block, this should return 2 values. But, it only returns 1 value.

src/module.model/oracle.price.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+}
+
+@Injectable()
+export class OraclePriceMapper {
+  public constructor (protected readonly database: Database) {
+  }
+
+  async getActive (timestamp: number): Promise<OraclePrice[] | undefined> {
```
jingyi2811, line 27: In the final block, this should return 2 values in an array

src/module.indexer/indexer.spec/oracle.price.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const timestamp = new Date().getTime()
+
+  const prices1 = [{ tokenAmount: '0.5@APPL', currency: 'EUR' }]
+  await client.oracle.setOracleData(oracleid1, timestamp, { prices: prices1 })
+
+  await container.generate(1)
+
+  const prices2 = [{ tokenAmount: '1.0@APPL', currency: 'EUR' }]
+  await client.oracle.setOracleData(oracleid2, timestamp, { prices: prices2 })
```
jingyi2811, line 65: Two oracle data are set

src/module.indexer/indexer.spec/oracle.price.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  await container.generate(1)
+  blockcount = await client.blockchain.getBlockCount()
+}
+
+describe('x', () => {
+  it('should wait for block height 1', async () => {
```
jingyi2811, line 72: This test is able to run successfully

src/module.indexer/indexer.spec/oracle.price.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const priceMapper = app.get(OraclePriceMapper)
+
+    let price = await priceMapper.get(`${oracleid1}-APPL-EUR`)
+    expect(price?.data.oracleid).toStrictEqual(oracleid1)
+    expect(price?.data.token).toStrictEqual('APPL')
+    expect(price?.data.currency).toStrictEqual('EUR')
+    expect(price?.data.amount).toStrictEqual('0.5')
+
+    price = await priceMapper.get(`${oracleid2}-APPL-EUR`)
```
jingyi2811, line 93: Two priceMapper object are set

src/module.model/oracle.price.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const OraclePriceMapping: ModelMapping<OraclePrice> = {
+  type: 'token_currency',
+  index: {
+    token_currency_timestamp: {
+      name: 'token_currency_timestamp',
+      partition: {
+        type: 'string',
+        key: (d: OraclePrice) => d.data.timestamp.toString()
+      }
```
monstrobishi, line 13: Since the prices above are set with the same timestamp, you are probably overwriting the original entry

src/module.indexer/indexer.spec/oracle.price.spec.ts
monstrobishi: Same timestamp, the entry is being overwritten because same key

monstrobishi:
@jingyi2811 @fuxingloh 

I think our indexer should take prices in the past 5 minutes rather than +/- 1 hour. Although this is how the RPC call does it, but I don't think this is correct behaviour, if we take prices in the past hour, an out of date oracle could skew the price substantially.

monstrobishi:
We don't need to follow the RPC call structure

src/module.indexer/indexer.spec/oracle.price.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+afterAll(async () => {
+  try {
+    await stopIndexer(app)
+  } finally {
+    await container.stop()
+  }
+})
+
+let oracleid1: string
+let oracleid2: string
```
monstrobishi, line 33: where is `oracleid2` set in the test?

src/module.indexer/indexer.spec/oracle.price.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  //
+  // it('should wait for block height 3', async () => {
+  //   await waitForHeight(app, blockcount)
+  //
+  //   const priceAggregrationMapper = app.get(OraclePriceAggregrationMapper)
+  //   const priceAggregation = await priceAggregrationMapper.get('APPL-EUR')
+  //
+  //   console.log(priceAggregation)
+  // })
+})
```
monstrobishi, line 97: would like to see a test with 3-4 different oracle ids and weightage

jingyi2811:
Close this PR. This PR was separated into 4 smaller PRs.
Will apply all the suggestions recommended into the PRs

PR #185, opened 2021-06-28
poolswap block aggregation

canonbrother:

/kind feature

PoolSwap block aggregation will be used to calculate APR and daily volume for DEX stats

Fixes part of #136


src/module.model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,63 @@
...
+    return await this.database.delete(PoolSwapAggregationMapping, id)
+  }
+}
+
+export interface PoolSwapAggregation extends Model {
+  /** date as id in format yyyy-mm-dd, eg: 2021-01-01 */
+  id: string
+
+  bucket: PoolSwapHourlyBucket
+}
```
fuxingloh, line 51: instead of this design, a better design would be having the bucket as part of the secondary sort key space

```ts
export interface PoolSwapAggregationBucket extends Model {
  poolId: string // partition key
  bucketId: string // sort key e.g. modded time, 10 minutes interval?
  
  // e.g. from TotalCount
  total: Bignumber
  count: number
}
```

Instead of querying a day, you can query the last 24 hours via range operator. (Slice Window)
If you want to query a day you just need to construct a 24-hour range query.

src/module.indexer/indexer.spec/poolswap.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  const aggregations = await mapper.query('1-0', 100)
+  console.log('aggregations: ', aggregations)
+
+  // indexed data get undefined
+  const agg1 = await mapper.get('1-0-2020-10-31T00:00')
+  console.log('agg1: ', agg1)
+
+  // manual data can get
+  const agg2 = await mapper.get('1-0-2020-08-31T19:20')
+  console.log('agg2: ', agg2)
```
canonbrother, line 114: @fuxingloh 
the manual added data is stored but indexed data

src/module.indexer/model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,110 @@
...
+        await this.mapper.put(testagg)
+
+        // check whether the aggregation above is stored into db - yes, but 'test' get undefined
+        const testget = await this.mapper.get(id)
+        console.log('testget: ', testget)
+
+        // query here is working correctly
+        // but query in test only return only the manual added data
+        const testquery = await this.mapper.query(poolId, 100)
+        console.log('testquery: ', testquery)
```
canonbrother, line 57: here checking whether the indexed and manual data is stored, both of them were stored can be `get` and `query`
but only manual data can be `get` and `query` in test.. super weird

src/module.model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface PoolSwapAggregation extends Model {
+  /**
+   * poolpair "1-0" as id
+   */
+  poolId: string
+  /**
+   * store datetime in ISO8601 string format YYYY-MM-DDTHH:mm:ss.sssZ, eg: 2021-04-01T13:20:00Z
+   */
+  bucketId: string // sort key e.g. modded time, 10 minutes interval?
```
fuxingloh, line 63: you should use number instead of a string, for better index compression and query performance. number are also easier to manage for indexes

src/module.indexer/model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,110 @@
...
+  // remove seconds and milliseconds
+  return dateTime.substr(0, dateTime.length - 8)
+}
+
+function constructId (poolId: string, bucketId: string): string {
+  return `${poolId}-${bucketId}`
+}
+
+function constructPoolId (fromTokenId: string, toTokenId: string): string {
+  return `${fromTokenId}-${toTokenId}`
```
fuxingloh, line 109: you should use pool token id instead of token pairs since in the future we will have pair with the same tokens

src/module.indexer/model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        aggregation.total = new BigNumber(aggregation.total).plus(data.fromAmount)
+        aggregation.count += 1
+
+        console.log('aggregation: ', aggregation)
+
+        await this.mapper.put(aggregation)
+
+        // manually add 1 here and test query
+        const testagg = PoolSwapAggregationIndexer.newPoolSwapAggregation(poolId, '2020-08-31T19:20')
```
fuxingloh, line 47:
```suggestion
        const testagg = PoolSwapAggregationIndexer.newPoolSwapAggregation(poolId, bucketId)
```

you are using a different id here than the `const id` hence you get nothing

canonbrother: no no.. here is to ensure the indexed data is stored against the test as test got undefined by getting indxed data

src/module.indexer/model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,92 @@
...
+
+        await this.mapper.put(aggregation)
+      }
+    }
+  }
+
+  async invalidate (block: RawBlock): Promise<void> {
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (!vout.scriptPubKey.asm.startsWith('OP_RETURN 4466547873')) {
```
monstrobishi, line 48: Should we separate out this constant somewhere? it's hard to understand what `4466547873` just by looking at the code

src/utils.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Generate random string
+ *
+ * @param {number} length of string
+ * @return {string}
+ */
+export function randomString (length: number): string {
+  const chars = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
+  let result = ''
+  for (let i = length; i > 0; --i) result += chars[Math.floor(Math.random() * chars.length)]
```
monstrobishi, line 10: Move this to a test utils, e.g. 

https://github.com/DeFiCh/jellyfish/blob/bf3e5bb2853ed2fff12f65c52deee4d16e562dd8/packages/jellyfish-transaction-builder/__tests__/test.utils.ts#L10

and look at implementing similar to the above, then base58 encode

src/module.indexer/indexer.spec/poolswap.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    // https://mainnet.defichain.io/#/DFI/mainnet/tx/c2a489adbebf97fb1dd0c695ee6d1f54eca4ea60165f12fd17f90923d6181d3e
+    // poolId: '0-2', fromAmount: 1100
+    asm: 'OP_RETURN 446654787317a914fe7d87680b758782401e747f8228cadc610b2047870000cc829c190000001976a91442511183ba47347712f9f3dca5d2d003b2e78d4888ac02ffffffffffffff7fffffffffffffff7f',
+    hex: '6a4c51446654787317a914fe7d87680b758782401e747f8228cadc610b2047870000cc829c190000001976a91442511183ba47347712f9f3dca5d2d003b2e78d4888ac02ffffffffffffff7fffffffffffffff7f'
+  }
+]
+
+function generateBlock (dummyScript: any, timestamp: number): any {
+  return {
+    hash: randomString(64),
```
fuxingloh, line 81: You should use the `Crypto` module instead of introducing utilities.

`crypto.randomBytes(32).toString('hex')`

src/module.model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+
+  /**
+   * poolpair "0-1" as id
+   */
+  poolId: string
+  /**
+   * store datetime in ISO8601 string format YYYY-MM-DDTHH:mm (minute scale), eg: 2021-04-01T13:20
+   */
+  bucketId: string
```
fuxingloh, line 63: I prefer using a number scale for timestamp instead of string for better sequential read performance.

Ignoring `id` since we will deprecate that in the future, we use it for compatibility reasons.

```
`2021-04-01T13:20` -> 16 bytes
`poolId` -> ` 1 byte
`total` -> `16 bytes`
`count` -> `8 bytes`
```

total = 41 bytes ish

vs

```
`number` -> 8 bytes
`poolId` -> ` 1 byte
`total` -> `16 bytes`
`count` -> `8 bytes`
```

total = 33 bytes ish

you can pack more sequential data into the disk chunk.

fuxingloh: secondly is non URL safe key is bad because you can't use it to construct URL query without encoding it

`ALPHA  DIGIT  "-" / "." / "_" / "~"`

fuxingloh: thirdly is time format standardization is not standardized everyone for everyone to decode, number is more straight forward

src/utils.ts
canonbrother: yea thanks..  i'm referring `crypto.randomBytes(32).toString('hex')`

src/module.model/poolswap.aggregation.ts
canonbrother: refer to https://github.com/DeFiCh/whale/pull/185#discussion_r661133261

src/module.model/poolswap.aggregation.ts
canonbrother: yea understand.. ady switched it to timestamp with rounded minutes

src/module.indexer/model/poolswap.aggregation.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function constructId (poolId: string, bucketId: number): string {
+  return `${poolId}@${bucketId}`
+}
+
+function constructPoolId (fromTokenId: string, toTokenId: string): string {
+  const sortedTokenIds = [Number(fromTokenId), Number(toTokenId)].sort((a, b) => a - b)
+  return `${sortedTokenIds[0]}-${sortedTokenIds[1]}`
+}
```
canonbrother, line 84: continue from https://github.com/DeFiCh/whale/pull/185#discussion_r660282563

src/module.indexer/model/poolswap.aggregation.ts
canonbrother: updated: https://github.com/DeFiCh/whale/pull/185#discussion_r661227359

src/module.indexer/model/poolswap.aggregation.ts
canonbrother: yea previously i didn't set poolpairId as that has no poolpairId in poolswap block
so that's quite handy to get the poolpairId,
```ts
tokenA = rpc.getToken(tokenA)
tokenB = rpc.getToken(tokenB)
poolpair = rpc.getPoolPair(tokenA-tokenB)
if !poolpair rpc.getPoolPair(tokenB-tokenA)
```

src/module.indexer/model/poolswap.aggregation.ts
canonbrother: understand.. hmm.. i think its better adding the desc there first for easier

src/module.indexer/model/poolswap.aggregation.ts
monstrobishi: This isn't a reliable way to get the poolpair id and it could fail in a lot of cases. The right way would be to search an existing index of poolpair info and find the poolpair where the tokenA and tokenB match up to those two

monstrobishi: A quick and dirty iteration on what you are doing above would be:
```
rpc.listpoolpairs().find(x => x.tokenA === tokenA && x.tokenB === tokenB) || (x.tokenB === tokenA && x.tokenA === tokenB))
```

but this is also not ideal as we want to know the state of that poolpair at that specific block height (incase anything got updated or changed). So we probably want to index createpoolpair and updatepoolpair custom tx and use those to store this information, then search as above within our own indexers.

Thoughts @fuxingloh

fuxingloh: @canonbrother means we need to do something similar to this https://github.com/DeFiCh/whale/pull/191

Where we index the status of a PoolPair (reward distribution), using the status to do calculations at each block height. Storing the latest block height, and always querying the latest for calculation.

monstrobishi:
@canonbrother are you still working on this PR? do we still need it or should we start a new one?

canonbrother:
> @canonbrother are you still working on this PR? do we still need it or should we start a new one?

@monstrobishi probably need start a new one.. its too outdated :sweat:

PR #188, opened 2021-06-28
Add `v0//blocks` & `v0//blocks/:id/transactions` APIs

### LINKED ISSUE #4: block.controller.ts

Block Controller for blocks.

`id` can be block hash or block height.

```
GET v1/:network/blocks
GET v1/:network/blocks/:id
GET v1/:network/blocks/:id/transactions
GET v1/:network/blocks/:id/transactions/:txId
GET v1/:network/blocks/:id/transactions/:txId/vins
GET v1/:network/blocks/:id/transactions/:txId/vouts
```

aikchun:

/kind feature

Fixes #4 

#### Additional comments?:


src/module.api/block.controller.ts
```diff
@@ -0,0 +1,15 @@
+import { Controller, Get } from '@nestjs/common'
+import { Block, BlockMapper } from '@src/module.model/block'
+
+@Controller('/v1/:network/block')
```
fuxingloh, line 4: it's v0 now

```suggestion
@Controller('/v0/:network/block')
```
```suggestion
@Controller('/v1/:network/block')
```

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+export class BlockController {
+  constructor (
+    protected readonly blockMapper: BlockMapper
+  ) {
+  }
+
+  @Get('/tip')
+  async getBlockTip (): Promise<Block | undefined> {
+    return await this.blockMapper.getHighest()
+  }
```
fuxingloh, line 14: don't need to use defid semantic

The block list operation should always be sorted in descending order hence `/blocks size = 1` should give you the same result.

src/module.api/block.controller.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+const container = new MasterNodeRegTestContainer()
+let controller: BlockController
+let app: TestingModule
+let mapper: BlockMapper
+
+beforeAll(async () => {
+  await container.start()
+  await container.waitForReady()
+  await container.waitForWalletCoinbaseMaturity()
+  const defidUrl = await container.getCachedRpcUrl()
```
fuxingloh, line 23: you should use e2e tests instead of spec.ts

it will be increasingly difficult to maintain spec for this scenario since you need to populate your own data.
Your own populated data is will break when upstream block model changed.

fuxingloh (changes requested):

src/module.api/block.controller.ts
aikchun: What do you mean by `defid semantic`?

fuxingloh: `getchaintips`, `tip` is a semantic of defid/bitcoind for whale since it's an indexer, it will index the best chain. You can just provide a normal RESTful design.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+    protected readonly blockMapper: BlockMapper,
+    protected readonly transactionMapper: TransactionMapper
+  ) {
+  }
+
+  @Get('/tip')
+  async getLatest (): Promise<Block | undefined> {
+    return await this.blockMapper.getHighest()
+  }
+
```
fuxingloh, line 17:
```suggestion

  @Get()
  async listBlocks(): Promise<...> {
    // pagination query block sort by latest
  }

```

aikchun: I have made changes for this.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+  constructor (
+    protected readonly blockMapper: BlockMapper,
+    protected readonly transactionMapper: TransactionMapper
+  ) {
+  }
+
+  @Get('/tip')
+  async getLatest (): Promise<Block | undefined> {
+    return await this.blockMapper.getHighest()
+  }
```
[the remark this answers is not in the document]
aikchun: This means we won't need the api endpoints for this?

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+  it('getBlock should get block with hash of block', async () => {
+    const blockHash = await container.call('getblockhash', [100])
+    const block = await controller.getBlock(blockHash)
+    expect(block?.height).toStrictEqual(100)
+    expect(block?.hash).toStrictEqual(blockHash)
+  })
+
+  it('getBlockTransactions should get transactions from a block by hash', async () => {
```
jingyi2811, line 39: Add test units to test negative flows => 
1) should throw error while getting non-existent hash
2) should listBlocks with an empty object if size ? next ? which is out of range

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+    return ApiPagedResponse.of(blocks, query.size, item => {
+      return item.id
+    })
+  }
+
+  @Get('/tip')
+  async getLatest (): Promise<Block | undefined> {
+    const blocks = await this.blockMapper.queryByHeight(1)
+    return blocks[0]
+  }
```
fuxingloh, line 30: remove

src/module.model/block.ts
```diff
@@ -36,7 +36,7 @@ export class BlockMapper {
     return blocks.length === 0 ? undefined : blocks[0]
   }
 
-  async queryByHeight (limit: number, lt?: number): Promise<Block[]> {
+  async queryByHeight (limit: number, lt?: ModelKey): Promise<Block[]> {
```
fuxingloh, line 39: revert, only number is allowed for each key

fuxingloh (changes requested):

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,27 @@
+import { Controller, Get, Param } from '@nestjs/common'
+import { Block, BlockMapper } from '@src/module.model/block'
+import { TransactionMapper, Transaction } from '@src/module.model/transaction'
+
+@Controller('/v0/:network/block')
```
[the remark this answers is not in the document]
aikchun: Made changes for this.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  @Get('/:id')
+  async getBlock (@Param('id') hash: string): Promise<Block | undefined> {
+    return await this.blockMapper.getByHash(hash)
+  }
+
+  @Get('/:id/transactions')
+  async getBlockTransactions (@Param('id') hash: string): Promise<Transaction[]> {
+    return await this.transactionMapper.queryByBlockHash(hash, 100)
+  }
```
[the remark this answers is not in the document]
aikchun: Made changes for this.

src/module.api/block.controller.ts
aikchun: Removed

src/module.model/block.ts
aikchun: Made changes.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @Get('/:id')
+  async getBlock (@Param('id') id: string): Promise<Block | undefined> {
+    const idIsNumber = isNumber(id)
+
+    if (idIsNumber) {
+      const height = parseInt(id)
+      return await this.blockMapper.getByHeight(height)
+    }
+    // id is not a number
```
fuxingloh, line 44:
```suggestion
```

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return await this.blockMapper.getByHash(id)
+  }
+
+  @Get('/:id/transactions')
+  async getBlockTransactions (@Param('id') hash: string, @Query() query: PaginationQuery): Promise<ApiPagedResponse<Transaction>> {
+    const transactions = await this.transactionMapper.queryByBlockHash(hash, query.size, query.next)
```
fuxingloh, line 51: should support query by height and number as well

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,57 @@
+import { Controller, Get, Param, Query } from '@nestjs/common'
+import { Block, BlockMapper } from '@src/module.model/block'
+import { TransactionMapper, Transaction } from '@src/module.model/transaction'
+import { ApiPagedResponse } from '@src/module.api/_core/api.paged.response'
+import { PaginationQuery } from '@src/module.api/_core/api.query'
+
+function isNumber (str: string): boolean {
+  return /^-?\d+$/.test(str)
+}
```
fuxingloh, line 9:
```suggestion
function isHeight (str: string): boolean {
  return /^[0-9]+$/.test(str)
}
```

no negative number and you should call it is height

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    return ApiPagedResponse.of(await this.blockMapper.queryByHeight(0), query.size, item => {
+      return item.id
+    })
+  }
+
+  @Get('/:id')
+  async getBlock (@Param('id') id: string): Promise<Block | undefined> {
+    const idIsNumber = isNumber(id)
```
fuxingloh, line 38: don't declare const that you don't plan to reuse

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+      const next = parseInt(query.next)
+      const blocks = await this.blockMapper.queryByHeight(query.size, next)
+      return ApiPagedResponse.of(blocks, query.size, item => {
+        return item.id
+      })
+    }
+
+    return ApiPagedResponse.of(await this.blockMapper.queryByHeight(0), query.size, item => {
+      return item.id
+    })
```
fuxingloh, line 33: structure this like how you others are structured, there shouldn't be an `if` statement

https://github.com/DeFiCh/whale/blob/1d3a4268be91c16d13a9955879e0ac43836fad8a/src/module.api/address.controller.ts#L86-L90

fuxingloh (changes requested):

src/module.api/block.controller.ts
aikchun: This `if` statement comes from the fact that `queryByHeight` requires a `number` for the second argument.

`query.next` is an optional string property which `parseInt` won't accept unless type guard is being applied.

Previously, I changed parameter type of the second argument to `ModelKey` from `number` to avoid this `if` statement [here](https://github.com/DeFiCh/whale/pull/188#discussion_r663465087)

Could you advise?

src/module.api/block.controller.ts
aikchun: Made changes.

src/module.api/block.controller.ts
aikchun: Removed the variable.

src/module.api/block.controller.e2e.ts
aikchun: Added tests as suggested.

src/module.api/block.controller.ts
fuxingloh: Check these out, do read them in full at the source:

https://github.com/DeFiCh/whale/blob/e0422e76a9796a068f7ebcc2ba8d3c0358c1ff4d/src/module.api/_core/api.paged.response.ts#L38-L40

https://github.com/DeFiCh/whale/blob/e0422e76a9796a068f7ebcc2ba8d3c0358c1ff4d/src/module.api/_core/api.paged.response.ts#L58-L60

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  @Get('/:id')
+  async getBlock (@Param('id') id: string): Promise<Block | undefined> {
+    if (isHeight(id)) {
+      const height = parseInt(id)
+      return await this.blockMapper.getByHeight(height)
+    }
+
+    // id is not a number
+    return await this.blockMapper.getByHash(id)
```
fuxingloh, line 44:
```suggestion
export function parseHeight (str?: string): number | undefined {
  if (str !== undefined && /^\d+$/.test(str)) {
    return parseInt(str)
  }
}

@Controller('/v0/:network/blocks')
export class BlockController {
  constructor (
    protected readonly blockMapper: BlockMapper,
    protected readonly transactionMapper: TransactionMapper
  ) {
  }

  @Get('')
  async listBlocks (
    @Query() query: PaginationQuery
  ): Promise<ApiPagedResponse<Block>> {
    const next: number | undefined = parseHeight(query.next)
    const blocks = await this.blockMapper.queryByHeight(query.size, next)

    return ApiPagedResponse.of(blocks, query.size, item => {
      return item.id
    })
  }

  @Get('/:id')
  async getBlock (@Param('id') id: string): Promise<Block | undefined> {
    const height = parseHeight(id)
    if (height !== undefined) {
      return await this.blockMapper.getByHeight(height)
    }

    // id is not a number
    return await this.blockMapper.getByHash(id)
```

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+@Controller('/v0/:network/blocks')
+export class BlockController {
+  constructor (
+    protected readonly blockMapper: BlockMapper,
+    protected readonly transactionMapper: TransactionMapper
+  ) {
+  }
+
+  @Get('')
+  async listBlocks (
```
canonbrother, line 22:
```suggestion
  async list (
```

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+  ): Promise<ApiPagedResponse<Block>> {
+    const height = parseHeight(query.next)
+    const blocks = await this.blockMapper.queryByHeight(query.size, height)
+    return ApiPagedResponse.of(blocks, query.size, item => {
+      return item.id
+    })
+  }
+
+  @Get('/:id')
+  async getBlock (@Param('id') id: string): Promise<Block | undefined> {
```
canonbrother, line 33: `getBlock` -> `get`

canonbrother:
```suggestion
  async get (@Param('id') id: string): Promise<Block | undefined> {
```

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+    expect(parseHeight('123')).toStrictEqual(123)
+  })
+
+  it('should return undefined for empty string', () => {
+    expect(parseHeight('')).toStrictEqual(undefined)
+  })
+
+  it('should return undefined for null', () => {
+    expect(parseHeight('')).toStrictEqual(undefined)
+  })
```
canonbrother, line 143: duplicated

src/module.api/block.controller.e2e.ts
canonbrother: consistency wise, can you follow our test pattern?
eg:
https://github.com/DeFiCh/whale/blob/main/src/module.api/poolpair.controller.e2e.ts

canonbrother (changes requested):
take note on code consistency

src/module.api/block.controller.e2e.ts
aikchun: Changed the test to input `undefined` in the argument

src/module.api/block.controller.ts
aikchun: Made these changes.

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+    const block = await controller.get('100')
+    expect(block?.height).toStrictEqual(100)
+  })
+
+  it('getBlockTransactions should get transactions from a block by hash', async () => {
+    const blockHash = await container.call('getblockhash', [100])
+    const paginatedTransactions = await controller.getBlockTransactions(blockHash, { size: 30, next: '10' })
+
+    // there's only one transaction
+    expect(paginatedTransactions.data.length).toStrictEqual(1)
```
jingyi2811, line 49: If there is only 1 transaction, we do not know whether the pagination will work or not.
Could you figure out how to increase the number of transactions?

jingyi2811:
You shouldn't commit and push `package-lock.json`

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+    expect(paginatedTransactions.data[0].block.height).toStrictEqual(100)
+  })
+
+  it('getBlockTransactions should get transactions from a block by height', async () => {
+    const paginatedTransactions = await controller.getBlockTransactions('100', { size: 30, next: '10' })
```
jingyi2811, line 54: Could we assume `getBlockTransactions` always return 0 or 1 record? 
If yes, you shouldn't add a pagination feature to `getBlockTransactions`
If no, you should think of a way to improve the test so `paginatedTransactions.data.length` returns more than 1 record. With more than 1 record, then only we can know whether the pagination can work or not.

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+  it('getBlockTransactions should get empty array when height is not valid', async () => {
+    const paginatedTransactions = await controller.getBlockTransactions('999999999999', { size: 30 })
+
+    expect(paginatedTransactions.data.length).toStrictEqual(0)
+  })
+
+  it('list should be able to get a paginated response of blocks', async () => {
+    const paginatedBlocks = await controller.list({ size: 30, next: '40' })
+
+    expect(paginatedBlocks.data.length).toStrictEqual(30)
```
jingyi2811, line 85: Write something like this to test whether pagination is working correctly or not

```
 const first = await controller.list({ size: 2 })

    expect(first.data.length).toStrictEqual(2)
    expect(first.page?.next).toStrictEqual('1')

    expect(first.data[0]).toStrictEqual(expect.objectContaining({ id: '0', symbol: 'DFI', symbolKey: 'DFI' }))
    expect(first.data[1]).toStrictEqual(expect.objectContaining({ id: '1', symbol: 'DBTC', symbolKey: 'DBTC' }))

    const next = await controller.list({
      size: 2,
      next: first.page?.next
    })

    expect(next.data.length).toStrictEqual(2)
    expect(next.page?.next).toStrictEqual('3')

    expect(next.data[0]).toStrictEqual(expect.objectContaining({ id: '2', symbol: 'DETH', symbolKey: 'DETH' }))
    expect(next.data[1]).toStrictEqual(expect.objectContaining({ id: '3', symbol: 'DBTC-DET', symbolKey: 'DBTC-DET' }))

    const last = await controller.list({
      size: 1,
      next: next.page?.next
    })

    expect(last.data.length).toStrictEqual(0)
    expect(last.page).toBeUndefined()
```

fuxingloh:
missing `whale-api-client` changes

src/module.api/block.controller.e2e.ts
aikchun: Unfortunately, I'm unable to deterministically generate a block with an arbitary set of transaction that I can test the pagination with.

This could be possible in the future with this https://github.com/DeFiCh/jellyfish/issues/294 and https://github.com/DeFiCh/jellyfish/pull/450.

Therefore may I suggest testing for the pagination for transactions be tested separately at a later date.

src/module.api/block.controller.e2e.ts
aikchun: I have made changes based on this.

I used `height` as a determining factor as the list is sorted by height and should be paged as so.

aikchun:
> I think you shouldn't commit and push `package-lock.json`

@jingyi2811 I have reverted `package-lock.json`.

jingyi2811:
> > I think you shouldn't commit and push `package-lock.json`
> 
> @jingyi2811 I have reverted `package-lock.json`.

Okat

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  it('list would return the latest set if next is undefined', async () => {
+    const paginatedBlocks = await controller.list({ size: 30 })
+
+    expect(paginatedBlocks.data.length).toStrictEqual(30)
+    expect(paginatedBlocks.data[0].height).toBeGreaterThanOrEqual(100)
+  })
+})
+
+describe('parseHeight', () => {
```
jingyi2811, line 147: We normally don't test the util function we create in whale.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+    protected readonly blockMapper: BlockMapper,
+    protected readonly transactionMapper: TransactionMapper
+  ) {
+  }
+
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<Block>> {
+    const height = parseHeight(query.next)
```
jingyi2811, line 25: Try something like this `const height = query.next? parseInt(query.next): undefined`
I think it offers the same functionality as parseHeight.

jingyi2811:
Cannonbrother mentioned `take note on code consistency`, which I agree too.
You probably may spend some time to read how the tests written by others

In general, you should only write 3 tests for function with pagination and 2 tests for function without pagination.

For example,

For list
- should list
- should list with pagination
- should list empty object as out of range

For get
- should get
- should throw error while getting non-existent token

For get with pagination
- should get
- should get with pagination
- should get empty object as out of range

The reason why we don't write test very detail in whale because we don't want to duplicate the test if it is already written in  the jellyfish.

jingyi2811:
Design your test like this

```

  describe('list', () => {
  }

  describe('get', () => {
  }

  describe('getBlockTransactions', () => {
  }

```

aikchun:
> Design your test like this
> 
> ```
> 
>   describe('list', () => {
>   })
> 
>   describe('get', () => {
>   })
> 
>   describe('getBlockTransactions', () => {
>   })
> ```

@jingyi2811 I have restructured the tests to look like how it is structured in other tests.

Also I have re written the test to test pagination and removed some unnecessary test.

However, I didn't remove the tests for utility function as I feel like it makes less sense to make the codebase less tested.

jingyi2811:
You push package-lock.json again

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @Get('/:id/transactions')
+  async getBlockTransactions (@Param('id') id: string, @Query() query: PaginationQuery): Promise<ApiPagedResponse<Transaction>> {
+    const height = parseHeight(id)
+
+    if (height !== undefined) { // if id is height
+      const block = await this.blockMapper.getByHeight(height)
+
+      if (block !== undefined) { // type guard
+        return ApiPagedResponse.of(await this.transactionMapper.queryByBlockHash(block.id, query.size, query.next), query.size, transaction => {
```
fuxingloh, line 51: This converges into call `queryByBlockHash`, your code must converge

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<Block>> {
+    const height = parseHeight(query.next)
+    const blocks = await this.blockMapper.queryByHeight(query.size, height)
+    return ApiPagedResponse.of(blocks, query.size, item => {
+      return item.id
```
fuxingloh, line 28:
```suggestion
      return item.height
```

fuxingloh (changes requested):

src/module.api/block.controller.ts
aikchun: Corrected this.

src/module.api/block.controller.ts
aikchun: Made changes to converge the code.

aikchun:
> You pushed package-lock.json again

I've reverted the changes properly this time. @jingyi2811

fuxingloh:
> missing `whale-api-client` changes

FYI you need to implement `whale-api-client` of the available endpoint you exposed.

packages/whale-api-client/src/api/block.ts
```diff
@@ -0,0 +1,64 @@
...
+    return await this.client.requestData('GET', `blocks/${id}`)
+  }
+
+  /**
+   * @param {string} id - hash or height of the block
+   * @param {number} [size=30] - size to query
+   * @param {string} [next] - next token for next slice of blocks
+   * @return {Promise<ApiPagedResponse<Transaction>>}
+   */
+  async getBlockTransactions (id: string, size = 30, next?: string): Promise<ApiPagedResponse<Transaction>> {
```
jingyi2811, line 32: size: number = 30

jingyi2811: Like how you wrote in line 14

packages/whale-api-client/__tests__/api/block.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  try {
+    await service.stop()
+  } finally {
+    await container.stop()
+  }
+})
+
+describe('get blocks', () => {
+  beforeAll(async () => {
+    await container.waitForWalletBalanceGTE(100)
```
jingyi2811, line 40: I think you can move this line to 27.

packages/whale-api-client/__tests__/api/block.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      vSize: expect.any(Number),
+      weight: expect.any(Number),
+      lockTime: expect.any(Number),
+      vinCount: expect.any(Number),
+      voutCount: expect.any(Number)
+    })
+    expect(transactions[0].block.height).toStrictEqual(37)
+  })
+
+  it('should getBlockTransactions through height', async () => {
```
jingyi2811, line 151: Need to add negative scenarios too

For eg: what happen if you call `await client.block.getBlockTransactions('$*@')`

packages/whale-api-client/__tests__/api/block.test.ts
```diff
@@ -0,0 +1,171 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
```
jingyi2811, line 1: Can you arrange your test like this

 ```
 describe('list', () => {
  })

  describe('get', () => {
  })

  describe('getBlockTransactions', () => {
  })
```

Kindly refer to other test file in the same folder

File: packages/whale-api-client/__tests__/api/block.test.ts [hunk 31]
```diff
@@ -0,0 +1,171 @@
...
+        size: expect.any(Number),
+        sizeStripped: expect.any(Number),
+        weight: expect.any(Number)
+      })
+    )
+    expect(first[0].height).toBeGreaterThanOrEqual(100)
+
+    const second = await client.paginate(first)
+    expect(second[0].height).toStrictEqual(first[29].height - 1)
+    expect(second.length).toStrictEqual(30)
```
jingyi2811, line 72: To verify pagination works correctly, we usually paginate 3 times

First, second and last

packages/whale-api-client/src/api/block.ts
```diff
@@ -0,0 +1,64 @@
...
+    return await this.client.requestData('GET', `blocks/${id}`)
+  }
+
+  /**
+   * @param {string} id - hash or height of the block
+   * @param {number} [size=30] - size to query
+   * @param {string} [next] - next token for next slice of blocks
+   * @return {Promise<ApiPagedResponse<Transaction>>}
+   */
+  async getBlockTransactions (id: string, size: number = 30, next?: string): Promise<ApiPagedResponse<Transaction>> {
```
fuxingloh, line 32:
```suggestion
  async getTransactions (id: string, size: number = 30, next?: string): Promise<ApiPagedResponse<Transaction>> {
```

packages/whale-api-client/src/whale.api.client.ts
```diff
@@ -60,6 +61,7 @@ export class WhaleApiClient {
   public readonly transactions = new Transactions(this)
   public readonly tokens = new Tokens(this)
   public readonly masternodes = new Masternodes(this)
+  public readonly block = new Blocks(this)
```
fuxingloh, line 64:
```suggestion
  public readonly blocks = new Blocks(this)
```

On second thought, I think to call it as `blocks` instead.

fuxingloh (changes requested):

packages/whale-api-client/src/api/block.ts
aikchun: Fixed.

packages/whale-api-client/src/whale.api.client.ts
aikchun: Fixed.

packages/whale-api-client/__tests__/api/block.test.ts
aikchun: Added assertion for the third and final page.

aikchun: Added assertion for the third and final page.

packages/whale-api-client/__tests__/api/block.test.ts
aikchun: Restructured the tests in this way.

packages/whale-api-client/__tests__/api/block.test.ts
aikchun: Added assertions for invalid input for `getTransactions` and `get`

packages/whale-api-client/__tests__/api/block.test.ts
aikchun: Made the changes.

packages/whale-api-client/src/api/block.ts
aikchun: Added type annotation for `size` parameter.

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,167 @@
...
+    expect(secondPage.data[0].height).toStrictEqual(firstPage.data[39].height - 1)
+
+    const lastPage = await controller.list({ size: 40, next: secondPage.page?.next })
+
+    expect(lastPage.data[0].height).toStrictEqual(secondPage.data[39].height - 1)
+    expect(lastPage.page?.next).toBeUndefined()
+  })
+
+  it('should return all the blocks if the size is out of range', async () => {
+    const paginatedBlocks = await controller.list({ size: 100000, next: '100' })
```
jingyi2811, line 94: Should fail if out of range

src/module.api/block.controller.e2e.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    expect(paginatedBlocks.data.length).toStrictEqual(100)
+    expect(paginatedBlocks.data[0].height).toBeGreaterThanOrEqual(99)
+  })
+
+  it('list would return the latest set if next is outside of range', async () => {
+    const paginatedBlocks = await controller.list({ size: 30, next: '100000' })
```
jingyi2811, line 101: Should fail if out of range

jingyi2811: Let say you have a list of objects => 9, 8, 7, 6, 5, 4, 3, 2, 1

list with next = 1  will give you error. Because there is no objects after it

src/module.api/block.controller.e2e.ts
aikchun: Why should this fail if the size is out of range?

It makes to sense to return all items if in a single page if the request page size is bigger is that data size.

By default, the size is 30, if for some reason, the chain has less than 30 blocks (or any other record for that matter), then we wouldn't return them any data?

Besides, it's part of the feature here https://github.com/DeFiCh/whale/blob/1d29ceeec4657e490f93a377bdf1cbe056216a01/src/module.api/poolpair.controller.e2e.ts#L88

src/module.api/block.controller.e2e.ts
aikchun: This shouldn't fail the request as we cannot assume that the client exactly how many blocks were already generated.

If `next` is too large we should return the first page so that the user is able to fetch the second page.

aikchun:
Added assertion for list block request when the `size` and `next` value are out of range.

src/module.api/block.controller.e2e.ts
jingyi2811: We add pagination feature to whale because we are unable to display every information in 1 page if the items are too many 

Let say there are 30 elements, and we want to display 10 in 1 page. There supposes to have 3 pages. 

1 ... 10, 11 ... 20, 21 ... 30

const first = await controller.list({ size: 10 }) will show 1 ... 10 elements
const second = await controller.list({ size: 10, next: 10 }) will show 11 ... 20 elements
const third = await controller.list({ size: 10, next: 20 }) will show 21 ... 30 elements
  
Next is always the last element of previous sublist

jingyi2811: Maybe you need to change the logic in `controller.list`

src/module.api/block.controller.e2e.ts
jingyi2811: I should write longer.`It makes to sense to return all items if in a single page if the request page size is bigger is that data size `=> This is correct.  When you search the word `out of range` in the project. You find this `it `

```
  it('should listTokens with an empty object if size 100 next 300 which is out of range', async () => {
    const result = await controller.list({ size: 100})

    expect(result.data.length).toStrictEqual(0)
    expect(result.page).toBeUndefined()
  })
```

Why don't you just copy the idea and combine your `size` and `next` tests into one?

jingyi2811:
I think this may help you

For queryByHeight, where the result is sorted by desc,
your next value should be lesser compare to previous sublist 

 ```
async queryByHeight (limit: number, lt?: number): Promise<Block[]> {
    return await this.database.query(BlockMapping.index.height, {
      limit: limit,
      order: SortOrder.DESC,
      lt: lt
    })
  }
```

For queryByBlockHash, where the result is sorted by asc,
your next value should be greater compare to previous sublist 

 ```
 async queryByBlockHash (hash: string, limit: number, gt?: string): Promise<Transaction[]> {
    return await this.database.query(TransactionMapping.index.block_txid, {
      partitionKey: hash,
      limit: limit,
      order: SortOrder.ASC,
      gt: gt
    })
  }
```

src/module.api/block.controller.e2e.ts
aikchun: It doesn't make sense to copy this test.

1. By combining `size` and `next` being out of range, doesn't elucidate the conditions that cause `results.data.length` to be 0. By testing the params individually, it could be shown that `next` being out of range is the primary cause for `result.data.length` to be 0.
2. listToken is fetching the list i [lexicographic order](https://github.com/DeFiCh/ain/blob/87a4760e06d286edcd481b810200b0b480e17282/src/masternodes/rpc_accounts.cpp#L241), which is an ascending order of sorts which makes sense if `next` is out of range, you are paging outside any datta.

list blocks, however, I refer back to @fuxingloh [comment](https://github.com/DeFiCh/whale/pull/188#discussion_r660259027) will always be in `descending` order therefore it makes that we will always have data even if the value for `next` is more than max height of the blocks.

aikchun:
> I think this may help you. In your test
> 
> For queryByHeight, where the result is sorted by desc,
> your next value should be lesser compare to previous sublist
> 
> ```
> async queryByHeight (limit: number, lt?: number): Promise<Block[]> {
>    return await this.database.query(BlockMapping.index.height, {
>      limit: limit,
>      order: SortOrder.DESC,
>      lt: lt
>    })
>  }
> ```
> 
> For queryByBlockHash, where the result is sorted by asc,
> your next value should be greater compare to previous sublist
> 
> ```
> async queryByBlockHash (hash: string, limit: number, gt?: string): Promise<Transaction[]> {
>    return await this.database.query(TransactionMapping.index.block_txid, {
>      partitionKey: hash,
>      limit: limit,
>      order: SortOrder.ASC,
>      gt: gt
>    })
>  }
> ```

list block operation will always in [descending order](https://github.com/DeFiCh/whale/pull/188#discussion_r660259027)

src/module.api/block.controller.e2e.ts
fuxingloh: Yeah @aikchun is right on this

src/module.api/block.controller.e2e.ts
aikchun: I have added tests to assertion for the length of data to be `0` when the `next` is `0` for both the controller and the `WhaleApiClient`.

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    const block = await this.parseHeightAndGetBlock(id)
+
+    const transactions = await this.transactionMapper.queryByBlockHash((block != null) ? block.id : id, query.size, query.next)
+
+    return ApiPagedResponse.of(transactions, query.size, transaction => {
+      return transaction.id
+    })
+  }
+
+  async parseHeightAndGetBlock (str?: string): Promise<Block | undefined> {
```
jingyi2811, line 50: `Str?` is optional. Can this be `str` ?

src/module.api/block.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    const block = await this.parseHeightAndGetBlock(id)
+
+    return block !== undefined ? block : await this.blockMapper.getByHash(id)
+  }
+
+  @Get('/:id/transactions')
+  async getBlockTransactions (@Param('id') id: string, @Query() query: PaginationQuery): Promise<ApiPagedResponse<Transaction>> {
+    const block = await this.parseHeightAndGetBlock(id)
+
+    const transactions = await this.transactionMapper.queryByBlockHash((block != null) ? block.id : id, query.size, query.next)
```
jingyi2811, line 43: Can this be `!== undefined`?

aikchun: Made changes for this

src/module.api/block.controller.ts
aikchun: Made the changes for this.

jingyi2811 (approved):

fuxingloh (approved):

PR #191, opened 2021-06-29
Index oracle weightage

### LINKED ISSUE #144: `Oracles` Service/Controller

jingyi2811:
What kind of PR is this?:
/kind feature

Index oracle weightage

Fixes #144

Additional comments?:


src/module.indexer/model/oracle.weightage.ts
```diff
@@ -0,0 +1,85 @@
...
+    super()
+  }
+
+  async index (block: RawBlock): Promise<void> {
+    const record: Record<string, OracleWeightage> = {}
+    const removedOracleIds: string[] = []
+
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (!vout.scriptPubKey.hex.startsWith('6a')) {
```
jingyi2811, line 19: I guess all custom transactions start with 6a.

src/module.indexer/model/oracle.weightage.ts
```diff
@@ -0,0 +1,85 @@
...
+    for (const aggregation of Object.values(record)) {
+      await this.mapper.put(aggregation)
+    }
+
+    for (const oracleId of removedOracleIds) {
+      await this.mapper.delete(oracleId)
+    }
+  }
+
+  async invalidate (block: RawBlock): Promise<void> {
```
jingyi2811, line 59: How to test this?

src/module.model/oracle.weightage.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Database } from '@src/module.database/database'
+
+const OracleWeightageMapping: ModelMapping<OracleWeightage> = {
+  type: 'oracle_weightage',
+  index: {
+    id: {
+      name: 'oracle_weightage_id',
+      partition: {
+        type: 'string',
+        key: (d: OracleWeightage) => d.id // oracleid
```
jingyi2811, line 12: id = oracleid

src/module.indexer/model/oracle.weightage.ts
```diff
@@ -0,0 +1,85 @@
...
+            const weightage = stack[1].tx.data.weightage
+            record[oracleId] = OracleWeightageIndexer.newOracleWeightage(oracleId, weightage)
+          }
+
+          if (stack[1]?.tx?.name === 'OP_DEFI_TX_REMOVE_ORACLE') {
+            const oracleId: string = stack[1].tx.data.oracleId
+            removedOracleIds.push(oracleId)
+          }
+        } catch (e) {
+          console.log(e)
```
jingyi2811, line 45: This is bad right?

src/module.indexer/model/oracle.weightage.ts
fuxingloh: atm there is no easy way to, I will refactor that in

src/module.model/oracle.weightage.ts
```diff
@@ -0,0 +1,41 @@
...
+    return await this.database.delete(OracleWeightageMapping, id)
+  }
+}
+
+export interface OracleWeightage extends Model {
+  id: string // oracleid
+  data: {
+    weightage: number
+  }
+}
```
fuxingloh, line 41: For Weightage, you also need to index it from a time-series data perspective. (This means you need to index the height). When you need to query the latest weightage, what you do is `getHighest()`. When a Weightage is removed you can mark it as the state: 'REMOVED' | 'LIVE'. This is because blockchain is immutable, and always time series. By doing so you can see a series of events that moves when the blockchain moves. Aka see the history of an oracle.

> For this scenario it's height as a time series variable. block height in an ever-increasing manner. similar to ScriptAggregationMapper with getLatest

Time series for reference:
1. https://en.wikipedia.org/wiki/Time_series
2. https://docs.microsoft.com/en-us/azure/architecture/data-guide/scenarios/time-series

fuxingloh (changes requested):

monstrobishi:
Maybe not in this PR, but I'd like to see some E2E tests for these oracle indexers

src/module.model/oracle.weightage.ts
monstrobishi: Just to be extra clear, use blockheight as index here not timestamp, then use in other PR: https://github.com/DeFiCh/whale/pull/177/files#r660674730

src/module.indexer/model/oracle.weightage.ts
```diff
@@ -68,14 +68,20 @@ export class OracleWeightageIndexer extends Indexer {
...
+    block: RawBlock,
     oracleId: string,
-    weightage: number
+    weightage: number,
+    state: WeightageStatus
   ): OracleWeightage {
     return {
       id: oracleId,
+      block: {
+        height: block.height
```
jingyi2811, line 79: Added this

src/module.indexer/model/oracle.weightage.ts
```diff
@@ -68,14 +68,20 @@ export class OracleWeightageIndexer extends Indexer {
...
+      },
       data: {
         weightage
-      }
+      },
+      state
```
jingyi2811, line 84: State can be LIVE or REMOVED

src/module.model/oracle.weightage.ts
```diff
@@ -20,6 +24,15 @@ export class OracleWeightageMapper {
   public constructor (protected readonly database: Database) {
   }
 
+  async getLatest (id: string): Promise<OracleWeightage | undefined> {
```
jingyi2811, line 27: Should I use hid? Like what was written in `script.aggregration.ts`

src/module.indexer/model/oracle.weightage.ts
```diff
@@ -27,28 +26,29 @@ export class OracleWeightageIndexer extends Indexer {
...
       }
     }
 
-    for (const aggregation of Object.values(record)) {
-      await this.mapper.put(aggregation)
-    }
+    for (const aggregation of Object.values(records)) {
+      // const latest = await this.mapper.getLatest(aggregation.id)
+      // if (latest !== undefined) {
+      //   aggregation.data.weightage = latest.data.weightage
```
jingyi2811, line 48: I added the `getLatest `function. It is working correctly. I test it by uncommenting this block of code.

src/module.indexer/indexer.spec/oracle.weightage.spec.ts
```diff
@@ -58,37 +58,39 @@ describe('Weightage - approveOracle', () => {
 
     const weight = await weightageMapper.get(oracleid)
```
jingyi2811, line 59: Missing `getLatest` interface in mapper class. If not I can add the test here

jingyi2811:
I think eventually I need to expose `getLatest` function to the endpoint.

Screenshot 2021-06-29 at 10 48 25 PM

src/module.model/oracle.weightage.ts
fuxingloh: you just need to provide a unique `id` this is for internal database tracking purposes, will be refactored out eventually

e.g. 
1. with just partition key = `id` as it's always unique
2. with both partition key and sort key, a unique `id` would be `partition key` + `sort key`

src/module.indexer/model/oracle.weightage.priceFeed.ts
```diff
@@ -0,0 +1,186 @@
...
+            for (let i = 0; i < priceFeeds.length; i += 1) {
+              const priceFeed = priceFeeds[i]
+
+              const token: string = priceFeed.token
+              const currency: string = priceFeed.currency
+
+              priceFeedRecords[`${oracleId}-${token}-${currency}`] = OracleWeightagePriceFeedIndexer.newOraclePriceFeed(block, oracleId, token, currency, OracleStatus.LIVE)
+            }
+          }
+
```
monstrobishi, line 51: we can use else ifs here

jingyi2811: Sure!

jingyi2811:
Added another mapper - Price Feed. (For eg: APPL / EUR, TESL / USD ...)

This is important because we want to know what is the distinct Price Feed available for every block. And, we use it to calculate the price

src/module.indexer/model/oracle.priceFeed.ts
```diff
@@ -0,0 +1,133 @@
...
+
+            const token: string = priceFeed.token
+            const currency: string = priceFeed.currency
+
+            priceFeedRecords[`${oracleId}-${token}-${currency}`] = OraclePriceFeedIndexer.newOraclePriceFeed(block, oracleId, token, currency, OraclePriceFeedStatus.LIVE)
+          }
+        } else if (stack[1]?.tx?.name === 'OP_DEFI_TX_UPDATE_ORACLE') {
+          // Removed priceFeed
+          const oracleId: string = stack[1].tx.data.oracleId
+          const allPriceFeeds = await this.mapper.getAll() ?? []
```
jingyi2811, line 44: Here, I try to do something like `select * from PriceFeed mapper where oracleId = blah blah blah`

This is what I wrote

I select all record
`
const allPriceFeeds = await this.mapper.getAll()`

Then, I select an individual record

`await this.mapper.getByOracleId(allPriceFeeds[y].id, oracleId)`

Seems silly. But it works. All the tests passed.

src/module.indexer/indexer.spec/oracle.priceFeed.spec.ts
```diff
@@ -0,0 +1,168 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
```
jingyi2811, line 1: Need to improve every test by comparing the final result with RPC calls. I shall add this later.

packages/whale-api-client/__tests__/stub.service.ts
```diff
@@ -34,4 +35,15 @@ export class StubService {
       expect(agg?.statistic.txCount).toStrictEqual(txCount)
     }, timeout)
   }
+
+  async waitForIndexedHeight (height: number, timeout: number = 30000): Promise<void> {
```
jingyi2811, line 39: Not sure I made a correct decision to add this

src/module.api/oracle.status.controller.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  describe('getAggregation', () => {
+    it('should getAggregation', async () => {
+      await waitForIndexedHeight(app, blockCount)
+
+      const result = await controller.getAggregation(oracleId)
+      expect(result?.data.weightage).toStrictEqual(2)
+    })
+
+    it('should return undefined if getAggregation with invalid id', async () => {
```
jingyi2811, line 56: I think this should throw an exception

src/module.model/oracle.status.ts
```diff
@@ -0,0 +1,59 @@
...
+    return await this.database.put(OracleStatusMapping, id)
+  }
+
+  async delete (id: string): Promise<void> {
+    return await this.database.delete(OracleStatusMapping, id)
+  }
+}
+
+export interface OracleStatus extends Model {
+  id: string // oracleid
```
fuxingloh, line 51: `id` must be unique, `oracleid` is the partition key, `height` is the sort key
hence the unique would be oracle id + height

if it's not unique, you will be override itself.

make you you setup your tests so that it overrides itself.

use `oracleId` semantic for consistency on whale side.

> wasn't well documented by me

fuxingloh (changes requested):

src/module.indexer/model/oracle.status.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (!vout.scriptPubKey.hex.startsWith('6a')) {
+          continue
+        }
+
+        const stack: any = toOPCodes(
+          SmartBuffer.fromBuffer(Buffer.from(vout.scriptPubKey.hex, 'hex'))
+        )
```
fuxingloh, line 25: check [0] is the correct signature

src/module.indexer/model/oracle.status.ts
```diff
@@ -0,0 +1,83 @@
...
+      for (const vout of txn.vout) {
+        const stack: any = toOPCodes(
+          SmartBuffer.fromBuffer(Buffer.from(vout.scriptPubKey.hex, 'hex'))
+        )
+
+        if (stack[1].tx.name === 'OP_DEFI_TX_APPOINT_ORACLE') {
+          const oracleId: string = txn.txid
+          const id = `${oracleId}-${block.height}`
+          await this.mapper.delete(id)
+        }
```
fuxingloh, line 61: you need to invalidate everything else too

packages/whale-api-client/src/api/oracle.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor (private readonly client: WhaleApiClient) {
+  }
+
+  /**
+   * Get current aggregated weightage of an oracle
+   *
+   * @param {string} id
+   * @return {Promise<OracleStatusAggregration>}
+   */
+  async getAggregation (id: string): Promise<OracleStatusAggregration> {
```
monstrobishi, line 16: This needs to be more self explanatory, infact nothing is being aggregated here, it's just Oracle status, so you can call this `getOracleStatus`

src/module.api/oracle.status.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+@Controller('/v0/:network/oracle/:id')
+export class OracleStatusController {
+  constructor (
+    protected readonly rpcClient: JsonRpcClient,
+    protected readonly deFiDCache: DeFiDCache,
+    protected readonly statusMapper: OracleStatusMapper
+  ) {
+  }
+
+  @Get('/aggregation')
```
monstrobishi, line 15: Call this oracle status

@fuxingloh what's the right convention for 2 word, is it kebab case

monstrobishi: or oracle/status

monstrobishi: ok I think just `status` works because it's behind the controller

src/module.model/oracle.status.ts
```diff
@@ -0,0 +1,59 @@
...
+@Injectable()
+export class OracleStatusMapper {
+  public constructor (protected readonly database: Database) {
+  }
+
+  async getLatest (id: string): Promise<OracleStatus | undefined> {
+    const aggregations = await this.database.query(OracleStatusMapping.index.id_height, {
+      partitionKey: id,
+      order: SortOrder.DESC,
+      limit: 100
```
canonbrother, line 32: 1 data only

src/module.model/oracle.status.ts
```diff
@@ -0,0 +1,59 @@
...
+import { OracleState } from '@whale-api-client/api/oracle'
+
+const OracleStatusMapping: ModelMapping<OracleStatus> = {
+  type: 'oracle_status',
+  index: {
+    id_height: {
+      name: 'oracle_status_id_height',
+      partition: {
+        type: 'string',
+        key: (d: OracleStatus) => d.id // oracleId - height
```
canonbrother, line 13:
```suggestion
        key: (d: OracleStatus) => d.id // oracleId + height
```

canonbrother (changes requested):

src/module.indexer/model/oracle.status.ts
jingyi2811: Sorry, I don't understand.  Below is a complete stack object for OP_DEFI_TX_APPOINT_ORACLE.

```
  console.log
    [
      OP_RETURN { type: 'OP_RETURN', code: 106 },
      OP_DEFI_TX {
        type: 'OP_DEFI_TX',
        tx: {
          signature: 1147556984,
          type: 111,
          name: 'OP_DEFI_TX_APPOINT_ORACLE',
          data: [Object]
        }
      }
    ]
```

The value of [0] is  OP_RETURN { type: 'OP_RETURN', code: 106 }

Should I validate the value?

src/module.model/oracle.status.ts
jingyi2811: I was testing something. Forget to switch back.

src/module.indexer/indexer.spec/oracle.status.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  it('should get weightage', async () => {
+    await setup()
+    await waitForHeight(app, height1)
+
+    const oracleStatusMapper = app.get(OracleStatusMapper)
+
+    const data1 = await oracleStatusMapper.get(`${oracleId}-${height1}`)
+    expect(data1?.data.weightage).toStrictEqual(1)
+    expect(data1?.state).toStrictEqual(OracleState.LIVE)
```
jingyi2811, line 101: If I appoint an Oracle and update it later, first mapper should has state = REMOVED right
If yes, what is the best way to doing that?

I actually think that `id = oracleId ` is better than `${oracleId}-${height1}`
Where I just need oracleId to update the original key

monstrobishi: No it should not have REMOVED, only the block that it gets removed

monstrobishi: We need the height so we know the weightage at that block height, when doing the price mapping

monstrobishi: @fuxingloh how do we handle dependencies in mapping? is there a possibility the price mapper runs before the oracle status mapper, and then the status is not available for weightage calculation?

jingyi2811: Screenshot 2021-07-02 at 5 34 30 PM 
 
I think they follow this order (See `_main.ts`)

monstrobishi: ok great, then we just need to order it correctly

packages/whale-api-client/src/api/oracle.ts
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * @param {string} id oracleId
+   * @return {Promise<OracleStatus>}
+   */
+  async getStatus (id: string): Promise<OracleStatus> {
+    return await this.client.requestData('GET', `oracle/${id}/status`)
+  }
+}
+
+export interface OracleStatus {
```
fuxingloh, line 21:
```suggestion
export interface OracleAppointed {
```

I think it's better to call it appointed since that's the action

fuxingloh (changes requested):

src/module.api/oracle.appointed.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { OracleAppointedMapper } from '@src/module.model/oracle.appointed'
+import { OracleAppointed } from '@whale-api-client/api/oracle'
+
+@Controller('/v0/:network/oracle')
+export class OracleAppointedController {
+  constructor (
+    protected readonly appointedMapper: OracleAppointedMapper
+  ) {
+  }
+
```
fuxingloh, line 11: `@Get('/appointments')`

query appointments?

src/module.api/oracle.appointed.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+  @Get('/:id/status')
+  async getStatus (
+    @Param('id') id: string
+  ): Promise<OracleAppointed | undefined> {
+    return await this.appointedMapper.getLatest(id)
+  }
```
fuxingloh, line 17: I think we just need to query appointments `list` instead of get `one`

packages/whale-api-client/__tests__/stub.service.ts
```diff
@@ -34,4 +35,15 @@ export class StubService {
...
+    const blockMapper = this.app?.get(BlockMapper)
+    if (blockMapper === undefined) {
+      throw new Error('StubService not initialized yet')
+    }
+    await waitForExpect(async () => {
+      const block = await blockMapper.getHighest()
+      await expect(block?.height).toBeGreaterThan(height)
+    }, timeout)
+  }
```
fuxingloh, line 48: I think you should create a PR for this separately and get it merged in early. Since this is a non-breaking change.

fuxingloh (changes requested):

src/module.api/oracle.appointed.controller.ts
```diff
@@ -0,0 +1,18 @@
+import { Controller, Get, Param } from '@nestjs/common'
+import { OracleAppointedMapper } from '@src/module.model/oracle.appointed'
+import { OracleAppointed } from '@whale-api-client/api/oracle'
+
+@Controller('/v0/:network/oracle')
+export class OracleAppointedController {
```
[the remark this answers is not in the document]
fuxingloh: to hold all oracles feature within this single controller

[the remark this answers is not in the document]
fuxingloh: actually, now that I think about it, if there isn't a use case for exposing status/appointed data I think you should just remove it entirely.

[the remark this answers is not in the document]
fuxingloh: since the weightage is just for internal calculation, nothing useful for external querying

src/module.model/oracle.appointed.ts
```diff
@@ -0,0 +1,48 @@
+import { Injectable } from '@nestjs/common'
+import { ModelMapping } from '@src/module.database/model'
+import { Database, SortOrder } from '@src/module.database/database'
+import { OracleAppointed } from '@whale-api-client/api/oracle'
+
+const OracleAppointedMapping: ModelMapping<OracleAppointed> = {
+  type: 'oracle_status',
+  index: {
+    oracleId_height: {
+      name: 'oracle_status_oracleId_height',
```
fuxingloh, line 10: btw to update all these `oracle_status` to `oracle_appointed`

monstrobishi: I would rather rename the controller to `oracle.status.ts` and keep status, oracle.appointed doesn't make much sense here

monstrobishi: Status is altered by appointed, updated and removed, so why name it based on only one of those triggers? And we are storing status here

fuxingloh: My beliefs are:

1. `Oracle` works by `Appointment` through Foundation or Operator Space
2. With an `Appointment ` it allows an owner/actor to perform price feed update
3. Triggers such as `Updates` changes `Appointment` of an assigned owner/actor

Hence the status is the reflection of the oracle appointment rather than anything else.

monstrobishi: Ok I guess that works as long as everything is consistent

src/module.api/oracle.appointed.controller.ts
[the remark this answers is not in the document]
jingyi2811: Ya, I also thought the same too.

packages/whale-api-client/__tests__/stub.service.ts
jingyi2811: Ok, sure!

jingyi2811:
Close this PR. I shall apply all the comments and suggestions to https://github.com/DeFiCh/whale/pull/201

PR #201, opened 2021-07-01
Index Oracle Price

### LINKED ISSUE #144: `Oracles` Service/Controller

jingyi2811:
What kind of PR is this?:
/kind feature

Index price aggregration

Fixes #144

Additional comments?:


src/module.model/oracle.priceData.ts
```diff
@@ -0,0 +1,73 @@
...
+@Injectable()
+export class OraclePriceDataMapper {
+  public constructor (protected readonly database: Database) {
+  }
+
+  async getActivePrices (token: string, currency: string, timestamp: number): Promise<OraclePriceData[] | undefined> {
+    return await this.database.query(OraclePriceDataMapping.index.token_currency_timestamp, {
+      partitionKey: `${token}-${currency}`,
+      order: SortOrder.ASC,
+      gte: timestamp - 300,
```
monstrobishi, line 42: Sorry disregard the earlier suggestion, let's keep this to 1 hour + 1 hour - same as C++ implementation

jingyi2811: Ok sure!

monstrobishi:
We should only need 2 indexers, not 4:
- Oracle Status, for status, price feed and weightage (this indexes AppointOracle, RemoveOracle and UpdateOracle TX)
- Price Data. for prices

jingyi2811:
It is fine for me.

packages/whale-api-client/src/api/oracle.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Get price for a token and currency
+   *
+   * @param {string} token
+   * @param {string} currency
+   * @return {Promise<OraclePriceAggregration>}
+   */
+  async getPrice (token: string, currency: string): Promise<OraclePriceAggregration> {
```
monstrobishi, line 56: we need a function that returns an array of prices across a time range with an interval.

E.g.

prices from 5th March to 10th March every 5 minutes (or blocks etc, if easier to implement)

monstrobishi (changes requested):

packages/whale-api-client/src/api/oracle.ts
monstrobishi: this is for drawing graphs

jingyi2811: Ok sure.  Maybe something like
`getPrices(timestampFrom: number, timestampTo: number, interval: number) `

jingyi2811: Also `async getPrice (token: string, currency: string)` is not enough
I need to add `async getPrice (token: string, currency: string, timestamp: number)` to query price at specific timestamp

packages/whale-api-client/src/api/oracle.ts
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Get price change percentage between 2 timestamps.
+   *
+   * @param {string} token
+   * @param {string} currency
+   * @param {number} timestamp1
+   * @param {number} timestamp2
+   * @return {Promise<BigNumber>}
+   */
+  async getPriceChangePercentage (token: string, currency: string, timestamp1: number, timestamp2: number): Promise<BigNumber> {
```
monstrobishi, line 72: is this needed for a specific page on the explorer?

monstrobishi:
This looks good, I'll approve once code coverage is improved

seems like the following have low coverage:

<html>
<body>


.../module.indexer/model/oracle.price.aggregration.ts | 68.25% <68.25%> (ø) |  
-- | -- | --
src/module.indexer/model/oracle.appointed.ts | 71.02% <71.02%> (ø) |  
src/module.indexer/model/oracle.price.data.ts | 78.12% <78.12%> (ø)


</body>
</html>

packages/whale-api-client/src/api/oracle.ts
jingyi2811: Jonathan removed the 24 hour price change percentage from his latest prototype
https://marvelapp.com/prototype/74de579/screen/80840940

I will confirm with him again

jingyi2811: Ok, this is what he told me
=> gathered feedback on it about it not being important for the user's using oracles

Okay, I shall remove the endpoint.

jingyi2811:
> This looks good, I'll approve once code coverage is improved
> 
> seems like the following have low coverage:
> 
> .../module.indexer/model/oracle.price.aggregration.ts	68.25% <68.25%> (ø)	 
> src/module.indexer/model/oracle.appointed.ts	71.02% <71.02%> (ø)	 
> src/module.indexer/model/oracle.price.data.ts	78.12% <78.12%> (ø)

All these files has an invalidate function, which is still unable to be tested.

Fuxing will refactor the whale code again. See his comment: https://github.com/DeFiCh/whale/pull/191#discussion_r660500874

fuxingloh (changes requested):

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  async index (block: RawBlock): Promise<void> {
+    const records: Record<string, OraclePriceAggregration> = {}
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (!vout.scriptPubKey.hex.startsWith('6a')) {
+          continue
+        }
+
```
ivan-zynesis, line 27: Should be checking vout type is `setOracleData` here? Otherwise the price will be no changes right? @monstrobishi please help to confirm.

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,126 @@
...
+          let sum = 0
+          let weightageSum = 0
+
+          for (let i = 0; i < priceDataResult.length; i += 1) {
+            const priceData = priceDataResult[i]
+
+            const weightageObj = await this.appointedWeightageMapper.getLatestByOracleIdHeight(priceData.data.oracleId, block.height)
+            const weightage = weightageObj?.data.weightage ?? 0
+
+            sum += priceData.data.amount * weightage
```
ivan-zynesis, line 53: BigNumber for precision.

src/module.indexer/model/oracle.price.aggregration.ts
monstrobishi: Actually the rest of the function makes no use of the `txn` or the `vout`, @fuxingloh can you help to take a look as I might be missing something but the forloop here seems unecessary? maybe even a separate indexer

monstrobishi: I think this code could live inside `appointedWeightageMapper`, or we can keep it as a separate indexer without the for loop above, but better to reduce the amount of indexers if we're not actually making use of the data in the block aside from block time and block height

monstrobishi: (By separate indexer I mean this indexer that I'm commenting on, not an entirely new one)

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,126 @@
...
+            tokenCurrenciesSet.add(`${m.data.token}-${m.data.currency}`)
+          })
+        }
+
+        for (const tokenCurrency of tokenCurrenciesSet) {
+          const data = tokenCurrency.split('-')
+          const token = data[0]
+          const currency = data[1]
+
+          const priceDataResult = await this.priceDataMapper.getActivePrices(token, currency, block.time) ?? []
```
ivan-zynesis, line 92: `this.priceDataMapper.getActivePrices` seems to query all the data in plus minus 1 hour (not sure 3600 is second or milli). This will potentially delete/invalidate `aggregation` belongs to older block?

monstrobishi: yes good spotting @ivan-zynesis !

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    for (const aggregation of Object.values(records)) {
+      await this.priceAggregrationMapper.put(aggregation)
+    }
+  }
+
+  async invalidate (block: RawBlock): Promise<void> {
+    const priceAggregrationIds: string[] = []
+
+    for (const txn of block.tx) {
```
ivan-zynesis, line 72: As (token, currency, block height/time) combination will be unique (based on my understanding on [L58](https://github.com/DeFiCh/whale/pull/201/files#diff-4b10347138343828d8ad3aeb7eb9b6b48313639f6166da974ba745a36408de7fR58)), no loop will be needed here, can invalidate everything (token currency pair) once for this block height.

src/module.indexer/model/oracle.price.aggregration.ts
monstrobishi: I see some potential issues but this only deletes the aggregated price at that block time, however block height <-> block time is not a strict mapping (some blocks can come before or after), so I see some potential issues here

monstrobishi: we may need to have another id in the key (e.g. block height) that is consistent for a specific block, so that we can support invalidation

src/module.indexer/model/oracle.price.aggregration.ts
monstrobishi: based on your comments I think we really need the unit tests for invalidation before this is live, as there are a lot of edgecases

src/module.indexer/model/oracle.price.data.ts
```diff
@@ -0,0 +1,189 @@
...
+                for (let k = 0; k < priceDataResult.length; k += 1) {
+                  const priceDataObj = priceDataResult[k]
+
+                  if (priceDataObj.state === OracleState.LIVE) {
+                    const token = priceDataObj.data.token
+                    const currency = priceDataObj.data.currency
+                    const height = priceDataObj.block.height
+                    const timestamp = priceDataObj.data.timestamp
+
+                    priceDataObj.state = OracleState.REMOVED
```
ivan-zynesis, line 58: Everything in oraclePriceData are occurred history, I wonder is this necessary.

src/module.indexer/model/oracle.price.aggregration.ts
jingyi2811: @monstrobishi Height is one of the keys for all mapper object

monstrobishi: ok then we should use height to invalidate here not time, it's more deterministic

src/module.indexer/model/oracle.appointed.ts
```diff
@@ -0,0 +1,373 @@
...
+            const oracleId: string = stack[1].tx.data.oracleId
+
+            // Update weightage
+            const oldWeightageObj = await this.appointedWeightageMapper.getLatestByOracleIdHeight(oracleId, block.height)
+            const oldHeight: number = oldWeightageObj?.block.height ?? 0
+
+            if (oldWeightageObj != null) {
+              oldWeightageObj.state = OracleState.REMOVED
+              weightageRecords[`${oracleId}-${oldHeight}`] = oldWeightageObj
+            }
```
monstrobishi, line 70: I feel we don't want to set REMOVED on the old data, as REMOVED represents the Oracle state being removed, it does not represent the Oracle was updated

src/module.indexer/model/oracle.appointed.ts
```diff
@@ -0,0 +1,373 @@
...
+            for (let i = 0; i < oldTokenCurrencyResult.length; i += 1) {
+              const oldTokenCurrencyObj = oldTokenCurrencyResult[i]
+
+              const oldToken: string = oldTokenCurrencyObj.data.token
+              const oldCurrency: string = oldTokenCurrencyObj.data.currency
+              const oldHeight: number = oldTokenCurrencyObj.block.height
+
+              oldTokenCurrencyObj.state = OracleState.REMOVED
+              tokenCurrencyRecords[`${oracleId}-${oldToken}-${oldCurrency}-${oldHeight}`] = oldTokenCurrencyObj
+            }
```
monstrobishi, line 87: Don't need to set REMOVED here

src/module.indexer/model/oracle.appointed.ts
```diff
@@ -0,0 +1,373 @@
...
+
+              if (priceDataObj.state === OracleState.LIVE) {
+                const token = priceDataObj.data.token
+                const currency = priceDataObj.data.currency
+                const height = priceDataObj.block.height
+                const timestamp = priceDataObj.data.timestamp
+
+                priceDataObj.state = OracleState.REMOVED
+
+                priceDataRecords[`${oracleId}-${token}-${currency}-${height}-${timestamp.toString()}`] = priceDataObj
```
monstrobishi, line 172: Don't need to set REMOVED here

src/module.indexer/model/oracle.appointed.ts
```diff
@@ -0,0 +1,373 @@
...
+          } else if (stack[1]?.tx?.name === 'OP_DEFI_TX_REMOVE_ORACLE') {
+            const oracleId: string = stack[1].tx.data.oracleId
+
+            // Remove weightage
+            const weightageObj = await this.appointedWeightageMapper.getLatestByOracleIdHeight(oracleId, block.height)
+            const height: number = weightageObj?.block.height ?? 0
+
+            if (weightageObj != null) {
+              weightageObj.state = OracleState.REMOVED
+              weightageRecords[`${oracleId}-${height}`] = weightageObj
```
monstrobishi, line 141: set weightage to 0 here

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+                  const price = prices[j]
+                  const currency: string = price.currency
+
+                  priceDataIds.push(`${oracleId}-${token}-${currency}-${block.height}-${timestamp.toString()}`)
+                }
+              }
+            }
+          }
+        }
+      }
```
monstrobishi, line 296: We are not invalidating `OP_DEFI_TX_REMOVE_ORACLE` here, need to handle that also

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,151 @@
...
+  ): Promise<OraclePriceAggregration | undefined> {
+    return await this.priceAggregrationMapper.getLatestByTokenCurrencyBlockTime(token, currency, timestamp)
+  }
+
+  @Get('/:token/:currency/:timestamp1/:timestamp2/:timeInterval/price/interval')
+  async getIntervalPrice (
+    @Param('token') token: string,
+      @Param('currency') currency: string,
+      @Param('timestamp1') timestamp1: number,
+      @Param('timestamp2') timestamp2: number,
```
monstrobishi, line 109: `timestamp1` -> `start`
`timestamp2` -> `end`

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+  OraclePriceData,
+  OracleState
+} from '@whale-api-client/api/oracle'
+import BigNumber from 'bignumber.js'
+
+@Injectable()
+export class OracleAppointedIndexer extends Indexer {
+  constructor (
+    private readonly appointedWeightageMapper: OracleAppointedWeightageMapper,
+    private readonly appointedTokenCurrencyMapper: OracleAppointedTokenCurrencyMapper,
```
monstrobishi, line 20: I think we can combine `OracleAppointedWeightageMapper` and `OracleAppointedTokenCurrencyMapper`

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+
+            for (let i = 0; i < priceFeeds.length; i += 1) {
+              const priceFeed = priceFeeds[i]
+
+              const token: string = priceFeed.token
+              const currency: string = priceFeed.currency
+
+              tokenCurrencyRecords[`${oracleId}-${token}-${currency}-${block.height}`] = OracleAppointedIndexer.newOracleAppointedTokenCurrency(block.height, oracleId, token, currency, OracleState.LIVE)
+            }
+          } else if (stack[1]?.tx?.name === 'OP_DEFI_TX_UPDATE_ORACLE') {
```
monstrobishi, line 60: `OP_DEFI_TX_APPOINT_ORACLE` and `OP_DEFI_TX_UPDATE_ORACLE` have very similar logic, see if you can combine it into another function or similar

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+                priceDataRecords[`${oracleId}-${token}-${currency}-${height}-${timestamp.toString()}`] = priceDataObj
+              }
+            }
+          } else if (stack[1]?.tx?.name === 'OP_DEFI_TX_SET_ORACLE_DATA') {
+            // NOTE(jingyi2811): Add price data
+            const timestamp: number = stack[1].tx.data.timestamp
+            const oracleId: string = stack[1].tx.data.oracleId
+            const tokens = stack[1].tx.data.tokens
+
+            for (let i = 0; i < tokens.length; i += 1) {
```
monstrobishi, line 183:
```suggestion
            tokens.forEach(token => {
```

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+              const token: string = tokens[i].token
+              const prices = tokens[i].prices
+
+              for (let j = 0; j < prices.length; j += 1) {
```
monstrobishi, line 187:
```suggestion
            prices.forEach(price => {
```

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+                const price = prices[j]
+
+                const currency: string = price.currency
+                const amount: number = price.amount
+
+                priceDataRecords[`${oracleId}-${token}-${currency}-${block.height}-${timestamp.toString()}`] = OracleAppointedIndexer.newOraclePriceData(block.height, oracleId, token, currency, new BigNumber(amount), timestamp, OracleState.LIVE)
```
monstrobishi, line 193:
```suggestion
priceDataRecords[`${oracleId}-${token}-${currency}-${block.height}-${timestamp.toString()}`] = 
  OracleAppointedIndexer.newOraclePriceData(
    block.height,
    oracleId,
    token,
    price.currency,
    new BigNumber(price.amount),
    timestamp,
    OracleState.LIVE)
```

monstrobishi: Try to structure the other lines like this, much easier to read. For the const vars above can just insert them directly like this

src/module.indexer/model/oracle.price.ts
```diff
@@ -0,0 +1,376 @@
...
+                // }
+              }
+            }
+          }
+        }
+      }
+    }
+
+    for (const record of Object.values(weightageRecords)) {
+      await this.appointedWeightageMapper.put(record)
```
monstrobishi, line 219: If we only need to store the values then we don't need the key in the objects above, they can just be arrays

packages/whale-api-client/__tests__/api/oracle.test.ts
```diff
@@ -0,0 +1,466 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { StubService } from '../stub.service'
+import { WhaleApiClient } from '../../src'
+import { StubWhaleApiClient } from '../stub.client'
+import BigNumber from 'bignumber.js'
+
+let container: MasterNodeRegTestContainer
```
monstrobishi, line 7: I suspect all of the describe sections should have their own container reference, otherwise you could face a race condition

src/module.indexer/indexer.spec/oracle.price.aggregration.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+} from '@src/module.indexer/indexer.spec/_testing.module'
+import { OraclePriceAggregrationMapper } from '@src/module.model/oracle.price.aggregration'
+import BigNumber from 'bignumber.js'
+
+const container = new MasterNodeRegTestContainer()
+let app: TestingModule
+
+describe('Price Aggregration - 1', () => {
+  beforeAll(async () => {
+    await container.start()
```
monstrobishi, line 17: We probably need individual container references here, since we are calling start multiple times on the same reference

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,116 @@
...
+      const token = data[0]
+      const currency = data[1]
+
+      // NOTE(jingyi2811): Calculate average price.
+      let sumBN = new BigNumber(0)
+      let weightageSum = 0
+
+      for (const oracleId of oracleIds) {
+        const weightageObj = await this.appointedWeightageMapper.getLatestByOracleIdHeight(oracleId, block.height)
+        const priceDataObj = await this.priceDataMapper.getLatestByOracleIdTokenCurrency(oracleId, token, currency, block.height)
```
ivan-zynesis, line 41: If I understand it correct ([indexer](https://github.com/DeFiCh/whale/pull/201/files#diff-a91ee61cc611bf57533a003129a5324bed0833e7fe205cfe5b20d20d02152889R77)), each `<oracle-id>-<token>-<currency>` can have more than 1 price tick in one block right? `getLatest...()` has limit 1 omits any older ticks.

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  static newOraclePriceAggregation (
+    height: number,
+    blockTime: number,
+    token: string,
+    currency: string,
+    amount: BigNumber
+  ): OraclePriceAggregration {
+    return {
+      id: `${token}-${currency}-${height}-${blockTime}`,
```
ivan-zynesis, line 104: `blockTime` doesnt seems necessary as part of id, since aggregated once per block.

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,116 @@
...
+      const priceDataResult = (await this.priceDataMapper.getActivePrices(token, currency, block.time) ?? [])
+
+      if (priceDataResult.length > 0) {
+        // NOTE(jingyi2811): Should include LIVE and REMOVED price data.
+        priceAggregrationIds.push(`${token}-${currency}-${block.height}-${block.time}`)
+      }
+    }
+
+    for (const id of priceAggregrationIds) {
+      await this.priceDataMapper.delete(id)
```
ivan-zynesis, line 92:
```suggestion
      await this.priceAggregrationMapper.delete(id)
```

src/module.indexer/model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,116 @@
...
+        tokenCurrenciesSet.add(`${m.data.token}-${m.data.currency}`)
+      })
+    }
+
+    for (const tokenCurrency of tokenCurrenciesSet) {
+      const data = tokenCurrency.split('-')
+      const token = data[0]
+      const currency = data[1]
+
+      const priceDataResult = (await this.priceDataMapper.getActivePrices(token, currency, block.time) ?? [])
```
ivan-zynesis, line 83: `indexer.invalidate()` is also running in forward manner. The priceData of this block most likely already deleted by now.
But make sense to make the main `this.indexer.reverse().forEach(idx => idx.invalidate()` instead

cc: @fuxingloh

ivan-zynesis (changes requested):

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+      @Param('currency') currency: string
+  ): Promise<OraclePriceAggregration | undefined> {
+    return await this.priceAggregrationMapper.getLatestByTokenCurrency(token, currency)
+  }
+
+  @Get('/:token/:currency/:timestamp/price')
+  async getPriceByTimestamp (
+    @Param('token') token: string,
+      @Param('currency') currency: string,
+      @Param('timestamp') timestamp: number
```
canonbrother, line 97: if not mistaken, you need parseInt on ts

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+    return await this.priceAggregrationMapper.getLatestByTokenCurrencyBlockTime(token, currency, timestamp)
+  }
+
+  @Get('/:token/:currency/:start/:end/:timeInterval/price/interval')
+  async getIntervalPrice (
+    @Param('token') token: string,
+      @Param('currency') currency: string,
+      @Param('start') start: number,
+      @Param('end') end: number,
+      @Param('timeInterval') timeInterval: number,
```
canonbrother, line 108: parseInt

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+    return ApiPagedResponse.of(sliceList, query.size, item => {
+      return item.timestamp.toString()
+    })
+  }
+
+  getSliceList<T> (list: T[], query: PaginationQuery, index: number): T[] {
+    let sliceList: T[] = []
+
+    if (query.next !== undefined) {
+      if (index >= 0) {
```
canonbrother, line 140: have this got `-ve` index??

src/module.model/oracle.appointed.token.currency.ts
```diff
@@ -0,0 +1,61 @@
...
+  index: {
+    id: {
+      name: 'oracle_appointed_token_currency_id',
+      partition: {
+        type: 'string',
+        key: (d: OracleAppointedTokenCurrency) => `${d.data.token}-${d.data.currency}-${d.block.height}`
+      }
+    },
+    oracleId_tokenCurrencyHeight: {
+      name: 'oracle_appointed_token_currency_oracleId-tokenCurrencyHeight',
```
canonbrother, line 17:
```suggestion
      name: 'oracle_appointed_token_currency_oracleId_tokenCurrencyHeight',
```

src/module.model/oracle.appointed.weightage.ts
```diff
@@ -0,0 +1,50 @@
+import { Injectable } from '@nestjs/common'
+import { ModelMapping } from '@src/module.database/model'
+import { Database, SortOrder } from '@src/module.database/database'
+import { OracleAppointedWeightage } from '@whale-api-client/api/oracle'
+
+const OracleAppointedWeightageMapping: ModelMapping<OracleAppointedWeightage> = {
+  type: 'oracle_appointed_weightage',
+  index: {
+    oracleId_height: {
+      name: 'oracle_appointed_weightage_oracleId-height',
```
canonbrother, line 10:
```suggestion
      name: 'oracle_appointed_weightage_oracleId_height',
```

src/module.model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,80 @@
+import { Injectable } from '@nestjs/common'
+import { ModelMapping } from '@src/module.database/model'
+import { Database, SortOrder } from '@src/module.database/database'
+import { OraclePriceAggregration } from '@whale-api-client/api/oracle'
+
+const OraclePriceAggregrationMapping: ModelMapping<OraclePriceAggregration> = {
+  type: 'oracle_price_aggregration',
+  index: {
+    tokenCurrency_heightBlockTime: {
+      name: 'oracle_price_aggregration_tokenCurrency-heightBlockTime',
```
canonbrother, line 10:
```suggestion
      name: 'oracle_price_aggregration_tokenCurrency_heightBlockTime',
```

src/module.model/oracle.price.aggregration.ts
```diff
@@ -0,0 +1,80 @@
...
+      limit: 1000000
+    })
+  }
+
+  async getLatestByTokenCurrencyBlockTime (token: string, currency: string, blockTime: number): Promise<OraclePriceAggregration | undefined> {
+    const data = await this.database.query(OraclePriceAggregrationMapping.index.tokenCurrency_blockTime, {
+      partitionKey: `${token}-${currency}`,
+      order: SortOrder.DESC,
+      gte: blockTime,
+      lte: blockTime,
```
canonbrother, line 62: hmm... weird.. is this working??

src/module.model/oracle.price.data.ts
```diff
@@ -0,0 +1,90 @@
+import { Injectable } from '@nestjs/common'
+import { ModelMapping } from '@src/module.database/model'
+import { Database, SortOrder } from '@src/module.database/database'
+import { OraclePriceData } from '@whale-api-client/api/oracle'
+
+const OraclePriceDataMapping: ModelMapping<OraclePriceData> = {
+  type: 'oracle_price_data',
+  index: {
+    oracleId_tokenCurrencyAmountHeight: {
+      name: 'oracle_price_data_oracleId-tokenCurrencyAmountHeight',
```
canonbrother, line 10:
```suggestion
      name: 'oracle_price_data_oracleId_tokenCurrencyAmountHeight',
```

src/module.model/oracle.price.data.ts
```diff
@@ -0,0 +1,90 @@
...
+        type: 'string',
+        key: (d: OraclePriceData) => d.data.oracleId
+      },
+      sort: {
+        type: 'string',
+        key: (d: OraclePriceData) => `${d.data.token}-${d.data.currency}-${d.data.amount.toString()}-${d.block.height}`
+      }
+    },
+    oracleIdTokenCurrency_height: {
+      name: 'oracle_price_data_oracleIdTokenCurrency-height',
```
canonbrother, line 21:
```suggestion
      name: 'oracle_price_data_oracleIdTokenCurrency_height',
```

src/module.model/oracle.price.data.ts
```diff
@@ -0,0 +1,90 @@
...
+        type: 'string',
+        key: (d: OraclePriceData) => `${d.data.oracleId}-${d.data.token}-${d.data.currency}`
+      },
+      sort: {
+        type: 'number',
+        key: (d: OraclePriceData) => d.block.height
+      }
+    },
+    tokenCurrency_timestamp: {
+      name: 'oracle_price_data_tokenCurrency-timestamp',
```
canonbrother, line 32:
```suggestion
      name: 'oracle_price_data_tokenCurrency_timestamp',
```

src/module.model/oracle.price.data.ts
```diff
@@ -0,0 +1,90 @@
...
+  index: {
+    oracleId_tokenCurrencyAmountHeight: {
+      name: 'oracle_price_data_oracleId-tokenCurrencyAmountHeight',
+      partition: {
+        type: 'string',
+        key: (d: OraclePriceData) => d.data.oracleId
+      },
+      sort: {
+        type: 'string',
+        key: (d: OraclePriceData) => `${d.data.token}-${d.data.currency}-${d.data.amount.toString()}-${d.block.height}`
```
canonbrother, line 17: the `interface OraclePriceData` should belong to here
those model interface should belong to model layer

canonbrother, line 17: is this how the `OraclePriceData.id` looks like??
can you move your model interface here? and note the id construction pattern?

canonbrother (changes requested):
too many duplicated fixture set up, can you refactor it?
will do another round review after cleanup...

src/module.api/oracle.controller.ts
```diff
@@ -0,0 +1,149 @@
...
+@Controller('/v0/:network/oracle')
+export class OracleController {
+  constructor (
+    protected readonly appointedTokenCurrencyMapper: OracleAppointedTokenCurrencyMapper,
+    protected readonly priceDataMapper: OraclePriceDataMapper,
+    protected readonly priceAggregrationMapper: OraclePriceAggregrationMapper
+  ) {
+  }
+
+  @Get('/token/currency')
```
canonbrother, line 26: @fuxingloh do we follow plural convention in api??

fuxingloh: hmm, yeah, we have it. I will refactor this PR

src/module.api/oracle.controller.ts
monstrobishi: I think take as string and parse to bignumber? Although unlikely for timestamp to need that precision, but on blockchain side it's 64 bit int

monstrobishi: Also do we want to standardise on milliseconds timestamp, because blockchain side uses seconds, I think controller should do milliseconds though

canonbrother: yea actually we don't really need to concern the timestamp precision as it has no decimal
but the `timestamp` came from API URL which is `string` but the usecase is `number`

```async getLatestByTokenCurrencyBlockTime (token: string, currency: string, blockTime: number):```

so far my review breakpoint here i'd ask for using `parseInt` which is part of nestjs controller feature

regarding the timestamp digit.. should standardise on milliseconds due to native js ts
but also need to check the length of timestamp, just in case, `if (ts.length === 10) ts * 1000`

fuxingloh:
Closed in favor of https://github.com/DeFiCh/whale/pull/281

> Significant refactor to move logic into indexer instead of Controller as the controller should not have any logic. Aggregation is thus also moved to price feed as each block height if there are any price feed.

PR #202, opened 2021-07-01
added poolpair stats `priceRatio` and `totalLiquidityUsd`

canonbrother:

/kind feature

Adding poolpair 'priceRatio' and 'totalLiquidityUsd' for explorer stat

Fixes part of #136


src/module.api/poolpair.service.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  async testPoolSwap (tokenFrom: string, tokenTo: string): Promise<string> {
+    const metadata: TestPoolSwapMetadata = {
+      tokenFrom: tokenFrom,
+      tokenTo: tokenTo,
+      from: this.testAddress,
+      to: this.testAddress,
+      amountFrom: 1
+    }
+    return await this.rpcClient.poolpair.testPoolSwap(metadata)
```
fuxingloh, line 64: I was wondering why you need this,

you don't need this to calculate the trading pair, you can just

> `'reserveA/reserveB': BigNumber | string`

find the pair DFI-USDT
where `a=DFI,b = USDT` via get poolpair

giving you the ratio

Because `testpoolpair` has reliance on consensus logic, where the above lookup doesn't allow for more agnostic replacement. (e.g. when we move to use whale index instead of reading from ain.)

fuxingloh (changes requested):

monstrobishi:
Unblocks https://github.com/DeFiCh/salmon/issues/77

src/module.api/poolpair.service.ts
canonbrother: oh.. i just realized.. updated!!

packages/whale-api-client/src/api/poolpair.ts
```diff
@@ -45,8 +45,15 @@ export interface PoolPairData {
     reserve: string // BigNumber
     blockCommission: string // BigNumber
   }
+  priceRatio: {
+    'tokenA/tokenB': string // BigNumber | string
+    'tokenB/tokenA': string // BigNumber | string
```
fuxingloh, line 50:
```suggestion
    ab: string // BigNumber | string
    ba: string // BigNumber | string
```

so that we can access it naturally without using `[]` for data access consistency across platform

packages/whale-api-client/__tests__/api/poolpair.test.ts
```diff
@@ -1,12 +1,17 @@
...
 import { createPoolPair, createToken, addPoolLiquidity, getNewAddress, mintTokens } from '@defichain/testing'
+import { PoolPairService } from '@src/module.api/poolpair.service'
+import { PoolPairData } from '@whale-api-client/api/poolpair'
+import BigNumber from 'bignumber.js'
 
 let container: MasterNodeRegTestContainer
 let service: StubService
 let client: WhaleApiClient
+let spy: jest.SpyInstance
+let poolPairService: PoolPairService
```
fuxingloh, line 14: for `.test.ts` and `e2e.ts` instead of mock spying how about creating a `DFI-USDT` pair? It only make sense to spy/mock on unitized testing

Since they are supposed to be e2e.

fuxingloh (changes requested):
2 Minor Changes, afterward I wanna do a minor refactoring PR into this branch to make it more resistant to code evolution

fuxingloh:
Reviewing this now, I might make changes to this PR directly

monstrobishi (approved):

fuxingloh (approved):

PR #215, opened 2021-07-06
Index customTx `createpoolpair` and `updatepoolpair`

canonbrother:

/kind feature

Index createpoolpair, and updatepoolpair.

Fixes part of #136


src/module.indexer/model/token.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (vout.scriptPubKey.asm.startsWith('OP_RETURN 4466547854')) { // 44665478 -> DFTX, 54 -> p -> create token
+          const stack: any = toOPCodes(
+            SmartBuffer.fromBuffer(Buffer.from(vout.scriptPubKey.hex, 'hex'))
+          )
+
+          const data: TokenCreate = (stack[1] as OP_DEFI_TX).tx.data
+
+          const token = await this.mapper.getLatest()
+          if (token !== undefined) {
+            const id = (Number(token.id) + 1).toString() // id increment
```
ivan-zynesis, line 28: Can't tell if there is any better (easy) way to retrieve this.
Currently implementation there is risk where normal token and LM token id jumble up

Indexer run single threaded, from one model to another model. I guess the token id assigned by defid should be increment chronologically, which can be creating new LM token then another normal token within the same block.

Might be no occurrence so far in the chain now. More than one token creation within one block can happen quite frequently when more product (stock, commodity etc) released, chances is there.

src/module.model/poolpair.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Database, SortOrder } from '@src/module.database/database'
+
+const PoolPairMapping: ModelMapping<PoolPair> = {
+  type: 'poolpair',
+  index: {
+    poolId: {
+      name: 'poolpair_poolId',
+      partition: {
+        type: 'string',
+        key: (p: PoolPair) => p.poolId
```
ivan-zynesis, line 13: If I understand how this ModelMapping work, `poolId` is also the unique token id right, then it will become one partition per poolpair and each partition has one row.

src/module.indexer/model/token.ts
canonbrother: yea i made a mistake here.. hmm... i should create another index the especially for `tokenId`

src/module.model/token.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Database, SortOrder } from '@src/module.database/database'
+
+const TokenMapping: ModelMapping<Token> = {
+  type: 'token',
+  index: {
+    height: {
+      name: 'token_height',
+      partition: {
+        type: 'number',
+        key: (t: Token) => t.block.height
```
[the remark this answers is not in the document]
canonbrother: yes for now.. its more on `sort` here

src/module.model/poolpair.ts
canonbrother: yea `poolId` is also the unique token, but rather i set `tokenA-tokenB` as primary key here

the reason is to adopt poolswap block logic
what will retrieve from poolswap block is `fromTokenId` and `toTokenId`
usage:
```
poolpairMapper.get(`${fromTokenId}-${toTokenId}`
```

canonbrother: its actually quite much tricky and workaround to structure the token and poolpair in LSM :sweat_smile:

src/module.indexer/model/token.ts
canonbrother: added `dctId` index

fuxingloh (changes requested):

fuxingloh:
I think instead of tracking DCT_ID increment via poolpair and token creation, we can just track `createpoolpair` and `updatepoolpair` DfTX and use the existing RPC protocol to find the actual `poolId`, this will cut down the number of indexes we have for now. 

Since RawBlock that you receive are already indexed in c++, the id should already be available. And you can use txid to get poolpair id.

The reason is for us to move so quickly to a full parity with c++ will be jarring as we will need to manage more state at whale. Increasing the complexity fold rather quickly, prefer to slowly increase to catch implementation issues.

src/module.indexer/model/poolpair.ts
```diff
@@ -0,0 +1,150 @@
...
+    private readonly tokenMapper: TokenMapper,
+    private readonly client: JsonRpcClient
+  ) {
+    super()
+  }
+
+  async index (block: RawBlock): Promise<void> {
+    for (const txn of block.tx) {
+      for (const vout of txn.vout) {
+        if (vout.scriptPubKey.asm.startsWith('OP_RETURN 4466547870')) { // 44665478 -> DFTX, 70 -> p -> create poolpair
```
jingyi2811, line 26: Initially I used number like `4466547870` But Fuxing asked me to change
See https://cakedefi.slack.com/archives/C01VBBJ8561/p1624344682274200?thread_ts=1624335293.268200&cid=C01VBBJ8561

Below is my latest code

```
    for (const txn of block.tx) {
      for (const vout of txn.vout) {
        if (!vout.scriptPubKey.hex.startsWith('6a')) {
          continue
        }

        const stack: any = toOPCodes(
          SmartBuffer.fromBuffer(Buffer.from(vout.scriptPubKey.hex, 'hex'))
        )

        if (stack[0]?.type === 'OP_RETURN' || stack[0]?.code === '106') {
          if (stack[1]?.tx?.name === 'OP_DEFI_TX_APPOINT_ORACLE') {
```

canonbrother: @jingyi2811  
its different as your previous implementation hardcoded the length
```
async index (block: RawBlock): Promise<void> {
  const records: Record<string, OraclePrice> = {}
  for (const txn of block.tx) {
    if (txn.hex.includes('6a414466547879')) {
      const buffer = SmartBuffer.fromBuffer(Buffer.from(txn.hex, 'hex'))
      const composable: any = new CSetOracleData(buffer)
```

jingyi2811: Ah, okay.

src/module.indexer/model/token.ts
```diff
@@ -25,17 +26,28 @@ export class TokenIndexer extends Indexer {
 
           const data: TokenCreate = (stack[1] as OP_DEFI_TX).tx.data
 
-          let dctId = await this.dctMapper.getLatest()
+          // get the latest tokenId via rpc listTokens
+          const tokens = await this.client.token.listTokens()
+          if (Object.keys(tokens).length === 0) {
```
ivan-zynesis, line 31: I think can reduce some codes here, effectively they behave the same.
```suggestion
          const token = tokens[data.symbol]
          if (token === undefined) throw new RpcNotFoundIndexerError('listTokens', data.symbol)
```

src/module.indexer/model/poolpair.ts
```diff
@@ -34,9 +34,10 @@ export class PoolPairIndexer extends Indexer {
             ? data.pairSymbol
             : await this.constructPairSymbol(data.tokenA.toString(), data.tokenB.toString())
 
-          const poolId = await this.newPoolId()
+          const rpcPoolPair = await this.client.poolpair.getPoolPair(pairSymbol)
```
ivan-zynesis, line 37:
```suggestion
          const rpcPoolPair = await this.client.poolpair.getPoolPair(pairSymbol)
          if (Object.keys(rpcPoolPair).length !== 1) throw
```
Can put a guard here, the indexer will make noise if same pair allowed in defid from different operator in future.

src/module.indexer/model/poolpair.ts
```diff
@@ -0,0 +1,150 @@
...
+          if (token === undefined || token.symbolId === undefined) {
+            throw new NotFoundIndexerError('index', 'UpdatePoolPair->Token', `${data.poolId}`)
+          }
+
+          const poolpair = await this.mapper.get(token.symbolId)
+          if (poolpair === undefined) {
+            throw new NotFoundIndexerError('index', 'UpdatePoolPair', `${data.poolId}`)
+          }
+
+          await this.mapper.delete(poolpair.id)
```
ivan-zynesis, line 114: Should be a major flow, cant be invalidating `PoolUpdatePair` deleting the poolPair entirely.
hmm, this requires some data structure (huge) changes to fix.

ivan-zynesis (changes requested):

src/module.indexer/model/token.ts
canonbrother: yea that's an ideal way but that has design issue, the resp is out of expected
```
{ '0': ...token, '1', ...token }
```

src/module.indexer/model/token.ts
canonbrother: solution was replaced by `rpc call`
https://github.com/DeFiCh/whale/pull/215#issuecomment-876188687

src/module.model/token.ts
```diff
@@ -0,0 +1,66 @@
+import { Injectable } from '@nestjs/common'
+import { Model, ModelMapping } from '@src/module.database/model'
+import { Database, SortOrder } from '@src/module.database/database'
```
fuxingloh, line 3: do we need tokens to be indexed? since you just need `module.model.poolpair.ts`

src/module.model/poolpair.ts
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
```
fuxingloh, line 14: Little confused about this index purpose, shouldn't it be

```suggestion
    poolId: {
      name: 'poolpair_poolId_height',
      partition: {
        type: 'string',
        key: (p: PoolPair) => p.poolId
      },
      sort: {
        type: 'height',
        key: (p: PoolPair) => p.block.height
      }
    }
```

for you to `getLatest(poolId)` information about the poolpair

canonbrother: because the usecase i was trying to make the query do this way
```ts
const poolpairs = await mapper.query() // list

const poolpair = await mapper.get('3') // get

const latest = await mapper.getLatest() // return latest poolpair
```
its a very simple design like block
so if set `sort` which the query will become
```ts
const poolpairs = await mapper.query('3') // return ??
```

src/module.model/token.ts
canonbrother: hmm.. because technically the poolpair is part of token
and i also think we will have token list design also
https://dex.defichain.com/mainnet/token

fuxingloh: I think because for now I rather we piggy back on ain for tokens API and only implement what we need.

src/module.model/poolpair.ts
fuxingloh: For the PoolPair indexes by right you should not be tracking `PoolPair` you should be tracking `PoolPairMutations` changes.

This means you track changes within PoolPair as it processes within the blockchain. `getLatest('poolId')` allow you to get the latest mutated change of that PoolPair to use for calculation of something else.

canonbrother: hmm.. @fuxingloh
In order to adopt poolswap index, the sources only container `fromTokenId` and `toTokenId`
so the poolpair index id will be constructed by tokenA-tokenB
while tracking `mutations`, then will have to store diff poolpair id, perhaps construct the id by tokenA-tokenB-{something}
then issue comes back to poolswap index, you got idea to get poolpair by id?? 

https://github.com/DeFiCh/whale/blob/539e9bed73c0736d908335a2f95146fd3f54c9d9/src/module.indexer/model/poolswap.aggregation.ts#L29-L42

fuxingloh: That would be `a-b-height` since height is mutating upwards.

You don't need to know something, you just need to get `a-b-` hence using getLatest() or getHighest()

For example `script.aggregation`, it uses the latest to get previous and build on top of it. This works because the ledger is always moving forward and getLatest is always the most recent.

https://github.com/DeFiCh/whale/blob/16d0f9813705dab2d9297f8301ee610c46dc3567/src/module.indexer/model/script.aggregation.ts#L58-L72

canonbrother: hmm. i put poolId as partition key currently, if that case then
```ts
// from
model {
  id: tokenAId-tokenBId // unique
  poolId: '2', // partition
  height: 22, // sort
}

// to
model {
  id: tokenAId-tokenBId-height
  symbolId:  tokenAId-tokenBId // partition
  height // sort
  poolId
}
```

fuxingloh: yup, this works, you can choose to use `poolId` or `tokenAId-tokenBId` 

depending on your use case, I think yours is `require getLatest(tokenAId-tokenBId)`?

canonbrother: yes use tokenAId-tokenBId is more straight forward due to poolswap object design, the current implementation is kind of tweaking around the workflow to set poolId as partitionKey :sweat_smile:

monstrobishi (changes requested):
Are you able to add invalidation tests for code coverage?

See
https://github.com/DeFiCh/whale/blob/main/src/module.indexer/indexer.spec/block.spec.ts#L60

and

https://github.com/DeFiCh/whale/blob/6680f57f595146bdb69bbaa0e336297a74b33acf/src/module.indexer/indexer.spec/oracle.appointed.token.currency.spec.ts#L247

src/module.indexer/error.ts
```diff
@@ -13,3 +13,21 @@ export class NotFoundIndexerError extends IndexerError {
...
+    super(`not found data from rpc ${action} ${item}`)
+  }
+}
+
+/**
+ * Result items.length validation on rpc READ
+ */
+export class RpcItemLengthError extends IndexerError {
+  constructor (item: string) {
+    super(`the ${item} length is not valid`)
```
monstrobishi, line 31: Cover these error conditions

src/module.indexer/model/poolpair.ts
```diff
@@ -0,0 +1,162 @@
...
+
+          const data: PoolCreatePair = (stack[1] as OP_DEFI_TX).tx.data
+
+          const symbol = data.pairSymbol !== ''
+            ? data.pairSymbol
+            : await this.constructPairSymbol(data.tokenA.toString(), data.tokenB.toString())
+
+          const rpcPoolPair = await this.client.poolpair.getPoolPair(symbol)
+
+          if (Object.keys(rpcPoolPair).length !== 1) throw new RpcItemLengthError('poolpair')
```
monstrobishi, line 36: Let's try to have brackets even on single line if statements otherwise it's error prone

src/module.indexer/model/poolpair.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (tokenA === undefined) {
+      throw new RpcNotFoundIndexerError('getToken', tokenAId)
+    }
+
+    const tokenB = await this.client.token.getToken(tokenBId)
+    if (tokenB === undefined) {
+      throw new RpcNotFoundIndexerError('getToken', tokenBId)
+    }
+
+    return `${tokenA[tokenAId].symbol}-${tokenB[tokenBId].symbol}`
```
monstrobishi, line 136: This isn't a reliable way to construct the symbol unfortunately, as it can be set to anything during `createpoolpair`

monstrobishi: If you need to find a certain poolpair containing tokens A and B, then loop through available poolpairs and check the tokenA and tokenB properties, or use a partition key of the tokens

monstrobishi, line 136: ignore previous comments, I see now this is just to mimic rpc default behaviour

src/module.indexer/indexer.spec/poolpair.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  await createPoolPair(container, 'ETH', 'DFI') // height 104
+  await createPoolPair(container, 'BTC', 'DFI')
+  await createPoolPair(container, 'USDT', 'DFI')
+  await createToken(container, 'DOGE')
+  await createPoolPair(container, 'DOGE', 'DFI')
+
+  await updatePoolPair(container, { pool: 'ETH-DFI', status: false, commission: 0.02 }) // height 109
+  await updatePoolPair(container, { pool: 'ETH-DFI', status: true, commission: 0.015 }) // height 110
+  await updatePoolPair(container, { pool: 'ETH-DFI', status: true, commission: 0.03 }) // height 111
+}
```
monstrobishi, line 55: Test poolpair with custom `pairSymbol`

src/module.indexer/model/poolpair.ts
canonbrother: i just see this comment after added the changes :sweat:

monstrobishi:
Closing this as we already have the functionality in main

PR #221, opened 2021-07-09
Fixed next param type in `token` and `poolpair` controllers

### LINKED ISSUE #220: Change `next` param type to string for `Tokens` and `PoolPairs` controllers
The type of `next` param is set to a number for `tokens` and `poolpairs` controllers which should be a string

```javascript
  @param {number} next set of PoolPairData
```

siradji:

/kind fix

`next` param on certain controllers is wrong.

Fixes #220


jingyi2811:
Screenshot 2021-07-09 at 8 51 05 PM
I think you still miss others

For eg: address.ts
Screenshot 2021-07-09 at 8 52 19 PM

fuxingloh (approved):

PR #224, opened 2021-07-10
Made collateral address optional if it is undefined. 

### LINKED ISSUE #206: `collateralAddress` return as undefined for PoolPair token

`undefined` data should not return the field, there are certain tokens type that return `undefined` data for certain fields. `collateralAddress` is one of the fields.

/triage accepted
/area module-api

siradji:

/kind fix

Removed undefined data from token controller 

Fixes #206

### DIFF
--- a/packages/whale-api-client/__tests__/api/tokens.test.ts
+++ b/packages/whale-api-client/__tests__/api/tokens.test.ts
@@ -34,7 +34,7 @@ describe('list', () => {
   it('should listTokens', async () => {
     const result = await client.tokens.list()
     expect(result.length).toStrictEqual(4)
-
+    expect(result[0].collateralAddress).toBeUndefined()
     expect(result[0]).toStrictEqual({
       id: '0',
       symbol: 'DFI',
@@ -55,8 +55,7 @@ describe('list', () => {
       destruction: {
         tx: '0000000000000000000000000000000000000000000000000000000000000000',
         height: -1
-      },
-      collateralAddress: ''
+      }
     })
   })
 
@@ -90,6 +89,7 @@ describe('list', () => {
 describe('get', () => {
   it('should get DFI by DFI numeric id', async () => {
     const data = await client.tokens.get('0')
+    expect(data.collateralAddress).toBeUndefined()
     expect(data).toStrictEqual({
       id: '0',
       symbol: 'DFI',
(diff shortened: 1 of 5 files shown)


src/module.api/token.controller.ts
```diff
@@ -74,6 +74,6 @@ function mapTokenData (id: string, tokenInfo: TokenInfo): TokenData {
     minted: tokenInfo.minted.toFixed(),
     creation: { tx: tokenInfo.creationTx, height: tokenInfo.creationHeight.toNumber() },
     destruction: { tx: tokenInfo.destructionTx, height: tokenInfo.destructionHeight.toNumber() },
-    collateralAddress: tokenInfo.collateralAddress
+    ...(!(tokenInfo.collateralAddress === '') && { collateralAddress: tokenInfo.collateralAddress })
```
fuxingloh, line 77:
```suggestion
    collateralAddress: tokenInfo.collateralAddress !== '' ? tokenInfo.collateralAddress : undefined
```

I think it's cleaner if you write it like this, instead of object restructuring and structuring.

fuxingloh (changes requested):

fuxingloh:
fix test to assert undefined for test

siradji:
> fix test to assert undefined for test

@fuxingloh , it is done.

packages/whale-api-client/__tests__/api/tokens.test.ts
```diff
@@ -34,7 +34,7 @@ describe('list', () => {
   it('should listTokens', async () => {
     const result = await client.tokens.list()
     expect(result.length).toStrictEqual(4)
-
+    expect(result[0].collateralAddress).toBeUndefined()
```
jingyi2811, line 37: In my opinion, you shouldn't check specifically if a value is undefined or not like in line `37`
We can assume the value is undefined if it is not exists in the result of the expect in line `38`

jingyi2811, line 37: In my opinion, you shouldn't check specifically if a value is undefined or not like in line 37
We can assume the value is undefined if it is not exists in the result of the expect in line 38

@fuxingloh what do you think?

fuxingloh: yeah jimmy

siradji: Done

jingyi2811 (approved):

fuxingloh (approved):

PR #270, opened 2021-07-27
global version prefixed endpoint

fuxingloh:

/kind chore

- Added global versioned prefixed endpoint: `/v0.0/regtest` `v{major}.{minor}/${network}`.
- Added automatically by `workflows/ci.yml` during release on client and server.
- Also updated nestjs to `>8` and various other related deps


monstrobishi (dismissed):
So does CI auto bump the version or do we manually increment? Otherwise it would break at every deployment right

fuxingloh:
> So does CI auto bump the version or do we manually increment? Otherwise it would break at every deployment right

Automatically for both client and server, I will validate the next release to make sure that's the case.

jingyi2811 (dismissed):

siradji (dismissed):

aikchun (approved):

PR #178, opened 2021-06-23
Add medianTime to all indexed model with block

ivan-zynesis:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
add medianTime to all indexer model which has "block", for UI to display time aside of block height

#### Which issue(s) does this PR fixes?:

Fixes #175

PR #281, opened 2021-07-30
Oracle Indexer

fuxingloh:

/kind feature

Refactor https://github.com/DeFiCh/whale/pull/201 with significant changes to move logic into indexer instead of Controller as the controller should not have any logic. Aggregation is thus also moved to price feed as each block height if there are any price feed.

- Completely remove database `scan` operations for performance reasons. With exceptions to aggregation.
- Both `Index` & `Invalidate` are tested
- [ ] More tests should be added for validity and conformation reasons

> Enhancement to this indexer should be limited to this scope, for future expansion, it should be implemented and properly tested in the jellyfish project for JS blockchain parity with C++.

#### Models Introduced:
- `PriceTicker`
- `Oracle`
- `OracleHistory`
- `OraclePriceFeed`
- `OraclePriceAggregated`
- `OracleTokenCurrency`


src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -0,0 +1,133 @@
...
+        oracles: {
+          active: aggregated.count,
+          total: oracles.length
+        }
+      },
+      currency: currency,
+      token: token,
+      id: `${token}-${currency}-${block.height}`,
+      key: `${token}-${currency}`,
+      sort: HexEncoder.encodeHeight(block.mediantime) + HexEncoder.encodeHeight(block.height)
```
monstrobishi, line 115: What's the rational for placing the medianTime before the height as a sort key?

src/module.api/prices.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+  async listPriceOracles (
+    @Param('key') key: string,
+      @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<OracleTokenCurrency>> {
+    const items = await this.oracleTokenCurrencyMapper.query(key, query.size, query.next)
+    return ApiPagedResponse.of(items, query.size, item => {
+      return item.oracleId
+    })
+  }
+}
```
monstrobishi, line 55: how do we provide the ability to display graphs on the front-end? would the paging possibly have the ability to add a `stride` (step size)?

src/module.indexer/model/dftx/appoint.oracle.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      for (const { token, currency } of data.priceFeeds) {
+        await this.oracleTokenCurrencyMapper.put({
+          id: `${token}-${currency}-${oracleId}`,
+          key: `${token}-${currency}`,
+          oracleId: oracleId,
+          token: token,
+          currency: currency,
+          weightage: data.weightage,
+          block: { hash: block.hash, height: block.height, medianTime: block.mediantime, time: block.time }
```
monstrobishi, line 50: I see a lot of use of median time through out, what's the rational behind that?

src/module.indexer/model/dftx/oracle.e2e.ts
```diff
@@ -0,0 +1,351 @@
...
+    await container.generate(1)
+    await waitForIndexedHeight(app, height)
+
+    {
+      const oracle = await app.get(OracleMapper).get(oracleId)
+      expect(oracle).toBeDefined()
+      const histories = await app.get(OracleHistoryMapper).query(oracleId, 10)
+      expect(histories.length).toStrictEqual(1)
+
+      const ta = await app.get(OracleTokenCurrencyMapper).query('TA-USD1', Number.MAX_SAFE_INTEGER)
```
monstrobishi, line 48: I really like the `TOKEN-CURRENCY` ID format, good thinking!

src/module.indexer/model/dftx/_abstract.ts
```diff
@@ -0,0 +1,16 @@
+import { RawBlock } from '@src/module.indexer/model/_abstract'
+import { blockchain } from '@defichain/jellyfish-api-core'
+import { DfTx } from '@defichain/jellyfish-transaction'
+
+export abstract class DfTxIndexer<T> {
+  abstract OP_CODE: number
+
+  abstract index (block: RawBlock, txns: Array<DfTxTransaction<T>>): Promise<void>
```
monstrobishi, line 8: The indexer includes a txn array now! Very handy!

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -0,0 +1,133 @@
...
+        continue
+      }
+      const key = `${token}-${currency}-${oracle.oracleId}`
+      const feeds = await this.feedMapper.query(key, 1)
+      if (feeds.length === 0) {
+        continue
+      }
+
+      // one hour -/+ time frame
+      if (Math.abs(feeds[0].time - block.time) < 3600) {
```
monstrobishi, line 90: We might want to grab the last element rather than `feeds[0]` as the last tx would be the one that gets ultimately processed on the ain side?

monstrobishi: Oh I see that it's sorted by `HexEncoder.encodeHeight(block.height) + txn.txid`, so above might be moot

monstrobishi: I think can ignore this, even in the worst case scenario where 2 prices are sent in the same block by the same oracle, the prices would likely end up the same

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -0,0 +1,133 @@
...
+import { Injectable } from '@nestjs/common'
+import { OraclePriceAggregated, OraclePriceAggregatedMapper } from '@src/module.model/oracle.price.aggregated'
+import { OraclePriceFeed, OraclePriceFeedMapper } from '@src/module.model/oracle.price.feed'
+import { HexEncoder } from '@src/module.model/_hex.encoder'
+import { OracleTokenCurrencyMapper } from '@src/module.model/oracle.token.currency'
+import BigNumber from 'bignumber.js'
+import { PriceTickerMapper } from '@src/module.model/price.ticker'
+
+@Injectable()
+export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
```
monstrobishi, line 13: Splitting the indexers by Dftx type makes the separtion of concerns much easier to understand

monstrobishi (approved):
Really amazing stuff, the only 2 things that stuck out were:
- I didn't fully understand why storing medianTime is necessary, although I'm sure it is, just lacking context
- Doesn't seem to have an easy way to obtain a range of prices over a certain time period with a specified interval (For graphing)

Overall nothing that would block, so I'm approving

src/module.indexer/model/dftx/appoint.oracle.ts
fuxingloh: the inclusion in `block: {}` is merely for displaying in UI as @ivan-zynesis requested it in another issue when he was doing wallet

src/module.api/prices.controller.ts
fuxingloh: Answered in another comment.

src/module.indexer/model/dftx/set.oracle.data.ts
fuxingloh: The `mediantime` being the `sortkey` allowing for graphing operations via querying `GT/LT`.

E.g. to get data within the last month you would query `LT today` and `GT today-1 month`. However, the index is not properly optimized as there is a lack of stride as you mention in another comment. For that to be implemented correctly, we need secondary indexes to re-aggregate the aggregated feed from oracles into time stride. 

> I think it would make sense to take this towards the jellyfish project for further expansion of the indices.

src/module.indexer/model/dftx/appoint.oracle.ts
fuxingloh: https://github.com/DeFiCh/whale/pull/178

jingyi2811:
I shall review this tomorrow

packages/whale-api-client/src/api/oracles.ts
```diff
@@ -1,9 +1,57 @@
...
+import { ApiPagedResponse } from '../whale.api.response'
 
 /**
  * DeFi whale endpoint for oracle related services.
  */
 export class Oracles {
   constructor (private readonly client: WhaleApiClient) {
   }
+
+  async list (size: number = 30, next?: string): Promise<ApiPagedResponse<Oracle>> {
```
jingyi2811, line 11: Not 50?
https://cakedefi.slack.com/archives/C01VBBJ8561/p1626415209080000?thread_ts=1626409403.079600&cid=C01VBBJ8561

src/module.api/prices.controller.ts
jingyi2811: I have the same question too. In my previous code, I have this function.

```
async getIntervalPrice (
    @Param('token') token: string,
      @Param('currency') currency: string,
      @Param('start') start: number,
      @Param('end') end: number,
      @Param('timeInterval') timeInterval: number,
      @Query() query: PaginationQuery
  ): Promise<ApiPagedResponse<OraclePriceInterval>> {
```

src/module.indexer/model/dftx/appoint.oracle.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      await this.oracleMapper.put({
+        id: oracleId,
+        weightage: data.weightage,
+        priceFeeds: data.priceFeeds,
+        block: { hash: block.hash, height: block.height, medianTime: block.mediantime, time: block.time }
+      })
+
+      await this.oracleHistoryMapper.put({
+        id: `${oracleId}-${block.height}-${txn.txid}`,
```
jingyi2811, line 34: I am thinking why this is not oracleId only

src/module.indexer/model/dftx/appoint.oracle.ts
```diff
@@ -0,0 +1,67 @@
...
+        sort: HexEncoder.encodeHeight(block.height) + txn.txid,
```
jingyi2811, line 35: What is the reason we encode this?

src/module.indexer/model/dftx/set.oracle.data.ts
jingyi2811, line 115: Could you explain a bit what this function does? `HexEncoder.encodeHeight`
Does it convert decimal to hexadecimal?

Screenshot 2021-08-01 at 11 01 53 PM

src/module.indexer/model/dftx/update.oracle.ts
```diff
@@ -0,0 +1,102 @@
...
+          block: previous.block
+        })
+      }
+    }
+  }
+
+  /**
+   * Get previous oracle before current height
+   */
+  private async getPrevious (oracleId: string): Promise<OracleHistory> {
```
jingyi2811, line 94: There are same getPrevious functions with same code in remove.oracle.ts and update.oracle.ts. Consider refactor it?

src/module.indexer/model/dftx/set.oracle.data.ts
monstrobishi: I think because it pads with 0000 it makes sorting more reliable e.g.

33
123

vs

0033
0123

jingyi2811:
I have spent time to write the tests to call the 

- AppointOracle
- RemoveOracle
- UpdateOracle
- setOracleData

rpcs and see if the output returned by LSM database match the output of the following rpcs

- listPrice
- listLatestRawPrice

 Should I add them to this PR?

src/module.indexer/model/dftx/set.oracle.data.ts
jingyi2811: Ah okay.

packages/whale-api-client/src/api/oracles.ts
fuxingloh: It's `60` https://github.com/DeFiCh/whale/blob/c1a5d441af71b89c27a5f293e2965cb076f1a4af/src/module.api/_core/api.query.ts#L15

src/module.indexer/model/dftx/appoint.oracle.ts
fuxingloh: id must be unique, thus oracle history requires `height` for `uniqueness`, but since oracle can be edited multiple time be block height,`txn.txid` further enforce the id uniqueness

src/module.indexer/model/dftx/appoint.oracle.ts
fuxingloh: Dealing with numbers and strings, you need to encode them in the same format allowing for lexicographic ordering.

src/module.indexer/model/dftx/update.oracle.ts
fuxingloh: it wasn't deterministically consistent when implemented to keep it more isolated for better defensive against regression.

for the larger refactoring effort of `module.indexer` it is planned for `Q4-Q1` with the learnings of mistakes and complexity endured by the team. above are all good questions, e.g. why height is encoded.

fuxingloh:
> I have spent time to write the tests to call the
> 
> * AppointOracle
> * RemoveOracle
> * UpdateOracle
> * setOracleData
> 
> rpcs and see if the output returned by LSM database match the output of the following rpcs
> 
> * listPrice
> * listLatestRawPrice
> 
> Should I add them to this PR?

Much appreciated!! But could you create a PR into this branch instead of committing into this branch directly.

fuxingloh:
Going to merge this in now, the changes targeting this branch can open PR to `main` instead.

PR #283, opened 2021-07-31
`oracle.price.aggregated.interval`

monstrobishi:

/kind feature

This uses timestamp to index prices based on interval.

Fixes #


src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -38,6 +57,12 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
       }
 
       await this.aggregatedMapper.put(aggregated)
+      for (const intervalMapper of this.intervalMappers) {
```
fuxingloh, line 60: very tricky to include it within `set oracle data`. for bucket aggregation, you need to aggregate the entire bucket to get the average.

if there isn't a `set oracle data` within that block interval then you also will not have it saved as a result.

another anomaly of this design is it's hard to query data for displaying your need since the block time isn't a good y-coordinate due to the sporadic nature. made worst with 1008 block adjustment now

monstrobishi: Thanks for the feedback. I believe since this is happening after aggregation so the only case it would display no data is if there is no set Oracle within +- 1 hour? If that is an issue I could move this into a separate indexer and query the latest aggregate price from the mapper there.

Yes I did have concerns about block height as (X?) coordinate. If you believe that to be an issue I could check using:

`timeNow > lastEntryMedianTime + interval`

fuxingloh: Speaking of which, for `PriceAggregation` if there isn't a `SetOracleData` it will not create a new Aggregated entry. This might be problematic.

A separate indexer makes sense where you query aggregate price and bucket aggregate it, keeping the separation of concern intact.

monstrobishi: Regarding just `PriceAggregation`, I think this should work as it is already. As long as it does not make any entries when not live, the front end should simply request the latest aggregated price to display (Regardless of how old it is).

For interval it should also be ok, the frontend will use time to plot the axis, so the areas where there are no entries will simply be represented as a straight line with no data points.

monstrobishi: I have modified the indexers to use time as interval

monstrobishi: However this makes testing slightly difficult, as we have to mock the blocktimes somehow (Unless the test runs for an hour). Perhaps I should just manually put blocks into the mapper.

monstrobishi: Would it make sense to do the following:
- Spin up a container and generate 100 blocks with one set oracle data per block
- List all blocks from block mapper
- Map all blocks and modify timestamp
- Start a new test app instance (or clear the existing block mapper)
- Pass in the modified list of blocks

packages/whale-api-client/src/api/prices.ts
```diff
@@ -23,6 +33,11 @@ export class Prices {
     return await this.client.requestList('GET', `prices/${key}/feed`, size, next)
   }
 
+  async getFeedWithInterval (token: string, currency: string, interval: PriceFeedTimeInterval, size: number = 30, next?: string): Promise<ApiPagedResponse<PriceFeed>> {
+    const key = `${token}-${currency}`
+    return await this.client.requestList('GET', `prices/${key}/${interval}/feed`, size, next)
```
fuxingloh, line 38:
```suggestion
    return await this.client.requestList('GET', `prices/${key}/feed/interval/${interval}`, size, next)
```

How about this

packages/whale-api-client/src/api/oracles.ts
```diff
@@ -55,3 +55,13 @@ export interface OraclePriceFeed {
...
+
+/**
+ * Time interval for graphing
+ */
+export enum OracleTimeInterval {
+  FIVE_MINUTES = 5 * 60,
+  TEN_MINUTES = 10 * 60,
+  ONE_HOUR = 60 * 60,
+  ONE_DAY = 24 * 60 * 60
+}
```
fuxingloh, line 67:
```suggestion
```

Probably don't need this?

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +67,15 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
     }
   }
 
+  private async indexIntervalMappers (block: RawBlock, token: string, currency: string, aggregated: OraclePriceAggregated): Promise<void> {
+    for (const intervalMapper of this.intervalMappers) {
+      const previous = await intervalMapper.query(`${token}-${currency}`, 1)
+      if (previous.length === 0 || (previous[0].block.medianTime + intervalMapper.interval) <= block.mediantime) {
+        await intervalMapper.put(aggregated)
```
fuxingloh, line 74: We can't just do this for the interval price, we need to average all prices within the interval slice.

```ts
if exist

price = existing.price * existing.count + new.price
count += 1

else

price = new.price
count = 1
```

fuxingloh (changes requested):

src/module.indexer/model/dftx/set.oracle.data.ts
monstrobishi: I've updated as per your suggestion

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +67,31 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
...
   }
 
+  private async indexIntervalMappers (block: RawBlock, token: string, currency: string, aggregated: OraclePriceAggregated): Promise<void> {
+    for (const intervalMapper of this.intervalMappers) {
+      const previous = await intervalMapper.query(`${token}-${currency}`, 1)
+      if (previous.length === 0) {
+        await intervalMapper.put(aggregated)
+      } else if ((previous[0].block.medianTime + intervalMapper.interval) <= block.mediantime) {
+        // Fetch all prices since the last interval
+        const prices = await this.aggregatedMapper.queryGt(`${token}-${currency}`, 5760, previous[0].sort)
```
ivan-zynesis, line 77: will this include the `previous[0]` row? Should the `prices.reduce(..., previous[0].price)` start with it?

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +67,31 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
...
+        if (prices.length !== 0) {
+          const averaged = prices.reduce((prev, curr): BigNumber => {
```
ivan-zynesis, line 79:
```suggestion
          const sum = prices.reduce((prev, curr): BigNumber => {
```

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +67,31 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
...
+            return prev.plus(curr.aggregated.amount)
+          }, new BigNumber(0))
```
ivan-zynesis, line 81: start with previous[0] price should make sense.

src/module.model/oracle.price.aggregated.interval.ts
```diff
@@ -0,0 +1,84 @@
...
+import { OraclePriceAggregated, OraclePriceAggregatedMapper } from './oracle.price.aggregated'
+
+export enum OracleIntervalSeconds {
+  FIVE_MINUTES = 5 * 60,
+  TEN_MINUTES = 10 * 60,
+  ONE_HOUR = 60 * 60,
+  ONE_DAY = 24 * 60 * 60
+}
+
+const getOraclePriceAggregatedIntervalMapping = (interval: number): ModelMapping<OraclePriceAggregated> => ({
```
fuxingloh, line 13: Instead of doing this why not have a single Model and Index by having the interval as part of the key.

It reduces the complexity of requiring us to create individual mapping.

fuxingloh (changes requested):

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +67,31 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
     }
   }
 
+  private async indexIntervalMappers (block: RawBlock, token: string, currency: string, aggregated: OraclePriceAggregated): Promise<void> {
+    for (const intervalMapper of this.intervalMappers) {
+      const previous = await intervalMapper.query(`${token}-${currency}`, 1)
+      if (previous.length === 0) {
+        await intervalMapper.put(aggregated)
+      } else if ((previous[0].block.medianTime + intervalMapper.interval) <= block.mediantime) {
+        // Fetch all prices since the last interval
```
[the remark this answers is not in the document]
ivan-zynesis: This is favorable, it spread out the workload. The largest interval has to do all the block when the last block come in.

@fuxingloh I think we do need definite starting point

block(t1), b(t2), b(t3), ..., b(t1 + 5min), b(t2 + 5min), ....

by doing this we will have 1 row per block, I think desired outcome is only have 1 row per interval.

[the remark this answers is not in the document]
monstrobishi: It is actually 1 row per interval right now, it only does put when interval is exceeded. But I get your point, the time slices should be based on the interval not when the block comes in. I'm wondering if we need a time-based indexing system (Schedular)? This would allow us to index things every X minutes rather than when a block comes in.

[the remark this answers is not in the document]
monstrobishi: Was replying to @ivan-zynesis just now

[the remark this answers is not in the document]
ivan-zynesis: > It is actually 1 row per interval right now

Yup, I did get that. Forward aggregate continuously as Fuxing suggested, could be quite tricky to keep track on the starting point.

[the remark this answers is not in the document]
fuxingloh: For script aggregation, it attempts to find a first. When the first isn't available it will create an empty record.

https://github.com/DeFiCh/whale/blob/bd8370ca648315bf6d6e1fd7193d6e1effa00a2d/src/module.indexer/model/script.aggregation.ts#L21-L27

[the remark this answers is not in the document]
monstrobishi: Let me clarify the forward aggregation method, you are referring to:
- Aggregate 5 minutes over all regular records
- Aggregate 10 minutes over the 5 minute records
- Aggregate 1 hour over the 10 minute records
- Aggregate 1 day over the 1 hour records

This is so that 1 day aggregation is not pulling up 2880 records correct?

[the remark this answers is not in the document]
fuxingloh: Nope,

Think of it from a data pipeline perspective ETL, MapReduce, etc

When new price aggregate comes in for 5 minutes pipeline.

1. Query previous in the 5 minutes index
2. Is the previous within the same 5 minutes bucket as now?
3. If so, aggregate forward via `count x price + price / count + 1`
4. Else, create new 5 minute bucket via `price=$, count=1`

When new price aggregate comes in for 10 minutes pipeline.

1. Query previous in the 10 minutes index
2. Is the previous within the same 10 minutes bucket as now?
3. If so, aggregate forward via `count x price + price / count + 1`
4. Else, create new 10 minute bucket via `price=$, count=1`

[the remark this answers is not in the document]
monstrobishi: Ok I got it.

This would require a new data structure and model for storing the aggregation buckets correct?

The actual end result will be stored in another model with the interval as index as you suggested.

canonbrother (dismissed):

src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -46,6 +61,46 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
     }
   }
 
+  private async indexIntervalMappers (block: RawBlock, token: string, currency: string, aggregated: OraclePriceAggregated): Promise<void> {
+    for (const interval of this.intervals) {
+      const previous = await this.aggregatedIntervalMapper.query(`${token}-${currency}-${interval}`, 2)
```
fuxingloh, line 66: don't understand why you need to query 2, you should be able to query the one before and 
- decide whether you need to create a new bucket
- or average between the old bucket to form a new bucket replacing the old bucket

fuxingloh (changes requested):

src/module.indexer/model/dftx/set.oracle.data.interval.ts
```diff
@@ -0,0 +1,148 @@
...
+  OracleIntervalSeconds,
+  OraclePriceAggregatedIntervalMapper
+} from '@src/module.model/oracle.price.aggregated.interval'
+import BigNumber from 'bignumber.js'
+import { mapPriceFeeds } from './set.oracle.data'
+
+@Injectable()
+export class SetOracleDataIntervalIndexer extends DfTxIndexer<SetOracleData> {
+  OP_CODE: number = CSetOracleData.OP_CODE
+  intervals: OracleIntervalSeconds[]
```
fuxingloh, line 16:
```suggestion
  intervals: OracleIntervalSeconds[] = [
      OracleIntervalSeconds.FIVE_MINUTES,
      OracleIntervalSeconds.TEN_MINUTES,
      OracleIntervalSeconds.ONE_HOUR,
      OracleIntervalSeconds.ONE_DAY
    ]
```

src/module.indexer/model/dftx/set.oracle.data.interval.ts
```diff
@@ -0,0 +1,148 @@
...
+    private readonly aggregatedIntervalMapper: OraclePriceAggregatedIntervalMapper
+  ) {
+    super()
+
+    this.intervals = [
+      OracleIntervalSeconds.FIVE_MINUTES,
+      OracleIntervalSeconds.TEN_MINUTES,
+      OracleIntervalSeconds.ONE_HOUR,
+      OracleIntervalSeconds.ONE_DAY
+    ]
```
fuxingloh, line 29:
```suggestion

```

src/module.indexer/model/dftx/set.oracle.data.interval.ts
```diff
@@ -0,0 +1,148 @@
...
+      key: `${token}-${currency}-${interval}`,
+      sort: aggregated.sort
+    })
+  }
+
+  async invalidateIntervalMapper (block: RawBlock, token: string, currency: string, aggregated: OraclePriceAggregated,
+    interval: OracleIntervalSeconds): Promise<void> {
+    const previous = await this.aggregatedIntervalMapper.query(`${token}-${currency}-${interval}`, 1)
+    // If count is 0 just delete
+    if (previous[0].aggregated.count === 0) {
```
fuxingloh, line 103:
```suggestion
    if (previous[0].aggregated.count === 1) {
```

mistake?

fuxingloh: should put this function below `invalidate`, easier to follow the flow

src/module.model/oracle.price.aggregated.interval.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  aggregated: {
+    amount: string
+    weightage: number
+    count: number
+    oracles: {
+      active: number
+      total: number
+    }
+  }
```
fuxingloh, line 66: should average `oracles.active/total` and `weightage` number for accuracy of interval data.

fuxingloh (changes requested):

fuxingloh:
Some conflicts to resolve for this.

fuxingloh (dismissed):

fuxingloh (approved):

Issue #246, opened 2021-07-15
Add `/stats` endpoint


fuxingloh:
#### What would you like to be added:

Provide an endpoint to get statistical data of the chain:

```
GET v0/:network/stats
```

return
```json
{
  "counts": {
    "masternode": 100,
    "block": 99,
    "token": 123,
    "price": 1,
    "oracle": 1
  },
  "tvl": {
    "total": 10234234324,
    "masternode": 1314134,
    "dex": 23434
  }
}
```

Endpoint data should be cached.

/triage accepted
/area module-api
/priority important-soon

PR #293, opened 2021-08-02
Added SemaphoreCache to limit concurrency for certain expensive operations

fuxingloh:

/kind chore

Certain operations such as #136 and #246 are expensive to compute, this will limit the concurrency of that critical zone to just 2 with double-checked locking.

This is a pre-req for #246.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@nestjs/platform-fastify": "^8.0.4",
         "@nestjs/schedule": "^1.0.0",
         "@nestjs/terminus": "^7.2.0",
+        "async-mutex": "^0.3.1",
         "bignumber.js": "^9.0.1",
         "cache-manager": "^3.4.3",
         "class-transformer": "^0.4.0",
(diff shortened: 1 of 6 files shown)


monstrobishi:
I did some reading on this recently, are we sending a `503 Retry-After` if the cache is empty but there is an ongoing calculation? Or is it able to block and return the cached result from the previous request.

fuxingloh:
> I did some reading on this recently, are we sending a `503 Retry-After` if the cache is empty but there is an ongoing calculation? Or is it able to block and return the cached result from the previous request.

This will block (45 seconds) until it acquires a lock, then process. When the lock is acquired it will check cache again before fetching. For double-checked locking.

fuxingloh:
> I did some reading on this recently, are we sending a `503 Retry-After` if the cache is empty but there is an ongoing calculation? Or is it able to block and return the cached result from the previous request.

This will block (45 seconds) until it acquires a lock, then process. When the lock is acquired it will check cache again before fetching. For double-checked locking.

monstrobishi (approved):
Thanks! According to `async-mutex` it will reject the promise in the case of a timeout, so in that case we could send a `503 Retry-After` to indicate to the client that they can try again after a set interval.

canonbrother (approved):

PR #317, opened 2021-08-04
Add more test units to validate Oracle indexers result with Oracle rpcs result

jingyi2811:

/kind feature

To ensure the result indexed by indexers accurate, we need to validate the values with values returned from Oracle rpcs like `getOracleData` and `listPrices`. 

Fixes #

### DIFF
--- a/src/module.indexer/model/dftx/appointed.oracle.spec.ts
+++ b/src/module.indexer/model/dftx/appointed.oracle.spec.ts
@@ -1,0 +1,111 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { createTestingApp, stopTestingApp, waitForIndexedHeight } from '@src/e2e.module'
+import { NestFastifyApplication } from '@nestjs/platform-fastify'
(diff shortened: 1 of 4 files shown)


src/module.indexer/model/dftx/appointed.oracle.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+import { JsonRpcClient } from '@defichain/jellyfish-api-jsonrpc'
+
+const container = new MasterNodeRegTestContainer()
+let app: NestFastifyApplication
+let client: JsonRpcClient
+
+beforeAll(async () => {
+  await container.start()
+  await container.generate(20)
+  await container.waitForWalletCoinbaseMaturity()
```
fuxingloh, line 17:
```suggestion
  await container.waitForWalletCoinbaseMaturity()
```

generate 20 is redundant because waitForWalletCoinbaseMaturity() supersede the block height to 100

jingyi2811: Haha, okay

PR #318, opened 2021-08-04
MainDfTxIndexer to log error instead of crashing to prevent indexer stall

fuxingloh:

/kind chore

Due to the sporadic nature of upgrading the blockchain, the DfTx schematic might change. This PR will prevent indexer stall by logging errors instead of erroring. 

> We can improve on this design by having a separated indexing pipeline. Where a failed pipeline won't affect another indexer pipeline.


monstrobishi (approved):
Test would be nice, but lgtm first

PR #320, opened 2021-08-05
Added paths to tsconfig to reference src rather than dist

### LINKED ISSUE #98: mono-repo tsconfig mapping for no build cross repo development

Update or add another `tsconfig` such that mono repo packages mappings are interlinked.

We could also export into the main directory instead of dist to have a cleaner export/import path, we need to update the packaging script for that.

#### Why is this needed:

After `npm install` when you need to run tests that required other packages to be built but isn't already, it causes the test to fail as it cannot find it in dist. When editing multiple packages without build, changes are not immediately reflected in the test as you need to build the entire mono repo again.

aikchun:

/kind feature

Fixes #98


packages/whale-api-client/src/index.ts
```diff
@@ -1,17 +1,17 @@
...
-export * as poolpairs from './api/poolpairs'
-export * as transactions from './api/transactions'
-export * as tokens from './api/tokens'
-export * as masternodes from './api/masternodes'
-export * as blocks from './api/blocks'
-export * as oracles from './api/oracles'
-export * as prices from './api/prices'
-export * as stats from './api/stats'
-export * as rawtx from './api/rawtx'
-export * as fee from './api/fee'
```
fuxingloh, line 14: this will be a breaking change for consuming libraries, instead of replacing it, add this additionally then add `@deprecated` comment to it.

tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "resolveJsonModule": true,
     "paths": {
       "@src/*": ["src/*"],
-      "@whale-api-client/api/*": ["packages/whale-api-client/src/api/*"]
```
fuxingloh, line 19: I remember why I didn't do this, it was to separate out `whale-api-client` with contents of `api/*`

The main issues for tsconfig mapping isn't `@src`, it's `whale-api-wallet` using `whale-api-client`.

packages/whale-api-client/src/index.ts
aikchun: Added back the export statements.

PR #323, opened 2021-08-05
Add basic APR calculation to `poolpair-controller`

### LINKED ISSUE #307: Add basic (no commission) APR calculation to poolpairs controller

Add basic (no commission) APR calculation to poolpairs controller

monstrobishi:

/kind feature

Add basic APR calculation to `poolpair-controller`

Fixes #307

### DIFF
--- a/packages/whale-api-client/src/api/poolpairs.ts
+++ b/packages/whale-api-client/src/api/poolpairs.ts
@@ -62,4 +62,8 @@ export interface PoolPairData {
     tx: string
     height: number
   }
+  apr: {
+    total: number
+    reward: number
+  }
 }
--- a/packages/whale-api-client/src/api/poolpairs.ts
+++ b/packages/whale-api-client/src/api/poolpairs.ts
@@ -62,8 +62,5 @@ export interface PoolPairData {
     tx: string
     height: number
   }
-  apr: {
-    total: number
-    reward: number
-  }
+  apr: { total: number, reward: number } | undefined
 }
--- a/packages/whale-api-client/src/api/poolpairs.ts
+++ b/packages/whale-api-client/src/api/poolpairs.ts
@@ -62,5 +62,8 @@ export interface PoolPairData {
     tx: string
     height: number
   }
-  apr?: { total: number, reward: number }
+  apr?: {
+    total: number // fractional number
+    reward: number // fractional number
+  }
 }
--- a/src/module.api/poolpair.controller.ts
+++ b/src/module.api/poolpair.controller.ts
@@ -8,59 +8,8 @@ import { PoolPairService } from './poolpair.service'
 import BigNumber from 'bignumber.js'
 import { PoolPairInfo } from '@defichain/jellyfish-api-core/dist/category/poolpair'
 
-@Controller('/poolpairs')
-export class PoolPairController {
-  constructor (
-    protected readonly rpcClient: JsonRpcClient,
-    protected readonly deFiDCache: DeFiDCache,
-    private readonly poolPairService: PoolPairService
-  ) {
-  }
-
-  /**
-   * @param {PaginationQuery} query
-   * @param {number} query.size
-   * @param {string} [query.next]
-   * @return {Promise<ApiPagedResponse<PoolPairData>>}
-   */
-  @Get('')
-  async list (
-    @Query() query: PaginationQuery
-  ): Promise<ApiPagedResponse<PoolPairData>> {
-    const result = await this.rpcClient.poolpair.listPoolPairs({
-      start: query.next !== undefined ? Number(query.next) : 0,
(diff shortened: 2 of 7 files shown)


jingyi2811 (approved):

src/module.api/poolpair.controller.ts
```diff
@@ -8,6 +8,44 @@ import { PoolPairService } from './poolpair.service'
 import BigNumber from 'bignumber.js'
 import { PoolPairInfo } from '@defichain/jellyfish-api-core/dist/category/poolpair'
 
+function mapPoolPair (id: string, info: PoolPairInfo, totalLiquidityUsd?: BigNumber,
```
fuxingloh, line 11: should leave the function below, since `@Controller` is contextually more important.

src/module.api/poolpair.service.ts
```diff
@@ -74,4 +74,40 @@ export class PoolPairService {
...
+      .plus(totalCustomRewards).toNumber()
+
+    // Total === reward right now as there is no volume information to calculate
+    // commission
+    const total = reward
+
+    return {
+      total,
+      reward
+    }
```
fuxingloh, line 111:
```suggestion
    return {
      reward: reward,
      total: reward
    }
```

src/module.api/poolpair.service.ts
```diff
@@ -74,4 +74,40 @@ export class PoolPairService {
       ttl: 180
     })
   }
+
+  async getDailyDFIReward (): Promise<BigNumber> {
+    const rpcResult = await this.rpcClient.masternode.getGov('LP_DAILY_DFI_REWARD')
+    return new BigNumber(rpcResult.LP_DAILY_DFI_REWARD)
```
fuxingloh, line 80: cache this function

src/module.api/poolpair.service.ts
```diff
@@ -74,4 +74,40 @@ export class PoolPairService {
...
+  }
+
+  async calculateAPRForPoolPair (info: PoolPairInfo): Promise<{ total: number, reward: number }> {
+    const dfiPriceUsdt: BigNumber = await this.getUSDT_PER_DFI() ?? new BigNumber(0)
```
fuxingloh, line 84: APR is in USDT?

shouldn't it be a percent relative to pool size?

fuxingloh (changes requested):

src/module.api/poolpair.service.ts
monstrobishi: It is a percentage, USD price is needed for rewards in usd over tvl in usd

monstrobishi: Now that you mention something looks off, it's missing the division! Good spotting

src/module.api/poolpair.controller.ts
```diff
@@ -92,6 +95,7 @@ export function mapPoolPair (id: string, info: PoolPairInfo, totalLiquidityUsd?:
     creation: {
       tx: info.creationTx,
       height: info.creationHeight.toNumber()
-    }
+    },
+    apr: apr ?? { total: 0, reward: 0 }
```
fuxingloh, line 99: should default to `undefined` and make it optional, similar to `totalLiquidity.usd` better if we throw error than display results wrongly.

fuxingloh (changes requested):

packages/whale-api-client/src/api/poolpairs.ts
```diff
@@ -62,4 +62,5 @@ export interface PoolPairData {
     tx: string
     height: number
   }
+  apr: { total: number, reward: number } | undefined
```
fuxingloh, line 65:
```suggestion
  apr?: { total: number, reward: number }
```

src/module.api/poolpair.service.ts
```diff
@@ -74,4 +74,39 @@ export class PoolPairService {
...
+      ttl: 180
+    })
+  }
+
+  async calculateAPRForPoolPair (info: PoolPairInfo): Promise<{ total: number, reward: number }> {
+    const dfiPriceUsdt = await this.getUSDT_PER_DFI() ?? 0
+
+    const totalCustomRewards = info.customRewards !== undefined
+      ? info.customRewards.reduce<string | BigNumber>((accum, customReward) => {
+        const [reward] = customReward.split('@')
```
fuxingloh, line 92: The right side of the split doesn't matter?

monstrobishi: The existing logic only handles DFI rewards (As it relies on DFI usd price, and rewards are expected to stay in DFI for the foreseable future).

I've added an assertion to make this more clear, and catch it if it becomes an issue in future.

fuxingloh:
- Refactor code smell
- Refactor types to be cleaner by referencing instead of declaring
- Certain errors should not fail but return undefined
- Should return fractional numbers instead of percentages 
- Updated ttl to 60 minutes since get daily dfi reward hardly ever changes

fuxingloh (approved):

PR #331, opened 2021-08-07
reconfigure probes to unhealthy if defid blocks are more than 4 headers behind

fuxingloh:

/kind fix

Reconfigure probes for dead if defid blocks are more than 4 headers behind.

`defid.readiness` dead if defid blocks are more than 4 headers behind
`model.readiness` dead synced blocks are more than 4 blocks behind

#### Additional comments?:

Also added probes healthy with peers testing with the new MasternodeGroup setup.

### DIFF
--- a/src/module.api/actuator.e2e.ts
+++ b/src/module.api/actuator.e2e.ts
@@ -1,29 +1,133 @@
-import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { MasternodeGroup, MasterNodeRegTestContainer } from '@defichain/testcontainers'
 import { NestFastifyApplication } from '@nestjs/platform-fastify'
 import { createTestingApp, waitForIndexedHeight } from '@src/e2e.module'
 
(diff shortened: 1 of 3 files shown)


jingyi2811:
Feel like reading spring boot actuator code :) So, 4 is better than 5?
https://github.com/DeFiCh/whale/blob/82bd37653b231a17377bf975971d35526d346613/src/module.model/_model.probes.ts#L34 You may need to change the figure 5 to 4 in the comment text too

fuxingloh:
> Feel like reading spring boot actuator code :) So, 4 is better than 5?
> https://github.com/DeFiCh/whale/blob/82bd37653b231a17377bf975971d35526d346613/src/module.model/_model.probes.ts#L34
> 
> You may need to change the figure 5 to 4 in the comment text too

Oh haha, it's inspired by spring boot implementation.

jingyi2811 (approved):

canonbrother (approved):

PR #344, opened 2021-08-17
`create.masternode` `resign.masternode` `block.minted` indexers

monstrobishi:

/kind feature

This adds indexing of masternodes to whale. The masternode controller has been replaced with the indexed data. EDIT: Ignore previous description, this works with masternode tx pre and post Eunos, there is just no timelock support here

Fixes #


src/module.indexer/model/dftx/create.masternode.ts
```diff
@@ -0,0 +1,61 @@
...
+
+      await this.masternodeMapper.put({
+        id: txn.txid,
+        sort: HexEncoder.encodeHeight(block.height) + txn.txid,
+        ownerAddress,
+        operatorAddress,
+        creationHeight: block.height,
+        resignHeight: -1,
+        mintedBlocks: 0,
+        timelock: 0,
```
canonbrother, line 44: something like this
```suggestion
        timelock: data.timelock || 0,
```

monstrobishi: there's no timelock in data

canonbrother: heyy?? wait...

monstrobishi: that's what your PR is for. but I suggest we get this merged in first. I can submit a new PR with timelock indexing

canonbrother: we will have `data.timelock` in block after eunospaya, so here suggest to check if data.timelock === undefined (before eunospay) set to zero, else follow the timelock from block

monstrobishi: there's literally no timelock variable in the data object. this is typescript

src/module.model/masternode.ts
```diff
@@ -0,0 +1,62 @@
...
+  async put (masternode: Masternode): Promise<void> {
+    return await this.database.put(MasternodeMapping, masternode)
+  }
+
+  async delete (id: string): Promise<void> {
+    return await this.database.delete(MasternodeMapping, id)
+  }
+}
+
+export interface Masternode extends Model {
```
canonbrother, line 44: i think we need mn.state here

monstrobishi: I had it originally then took it out. mn state should be inferred from creation and resign height rather than indexed

monstrobishi: See the controller for more details

canonbrother: some more TODO for my own about the tm1,2,3,4 after eunospaya mn merge
https://mydeficha.in/masternodes.php

canonbrother (approved):

canonbrother (dismissed):

src/module.indexer/model/dftx/create.masternode.ts
```diff
@@ -0,0 +1,61 @@
...
+      let operatorAddress = ownerAddress
+
+      // This is actually the operatorPubKeyHash but jellyfish deserializes like so
+      if (data.collateralPubKeyHash !== undefined) {
+        if (data.type === MasternodeKeyType.PKHashType) {
+          operatorAddress = P2PKH.to(this.network, data.collateralPubKeyHash).utf8String
+        } else { // WitV0KeyHashType
+          operatorAddress = P2WPKH.to(this.network, data.collateralPubKeyHash).utf8String
+        }
+      }
```
fuxingloh, line 34: hmmm, the other Indexed data like `ScriptActivity` we keep it network agonistic, maybe it makes more sense to resolve them at query time instead of storing them at the resolved state

> Up to you if you want to make the change, but this PR is good. Since we are going to consolidate jellyfish sub-projects into one, we can change them in the future.

fuxingloh (changes requested):
Looks good to me, I requested the change because with `jellyfish:0.39.0` merged, there is a breaking change as webster fixed the issue you mentioned about it being the wrong field. in `create.masternode.ts`

The DfTx has also been renamed to `CreateMasternode` and `ResignMasternode` with the capital `N

src/module.indexer/model/dftx/create.masternode.ts
monstrobishi: Ok so I can store it as pubkeyhash and resolve in the service. That sounds good. Script resolution shouldn't be a huge performance bottleneck on request right? No need to cache?

fuxingloh: Nope, we don't need to cache, it's a fairly simple deterministic code. Way faster than `JSON.stringify()`

monstrobishi: @fuxingloh I think I will stick to indexing the address as for operator I do have `data.operatorType` to go off but for owner I'm using `txn.vout[1].scriptPubKey.addresses[0]`.

Otherwise assuming `txn.vout[1].scriptPubKey.hex` is the pubkeyhash equivalent, is there a straightforward way to convert the owner pub key into an address without knowing the type down the line?

monstrobishi:
Will add timelock and tests here too

packages/whale-api-client/__tests__/api/masternodes.test.ts
```diff
@@ -28,34 +33,36 @@ afterAll(async () => {
...
-      mintedBlocks: 0,
-      owner: { address: 'bcrt1qyeuu9rvq8a67j86pzvh5897afdmdjpyankp4mu' },
-      operator: { address: 'bcrt1qurwyhta75n2g75u2u5nds9p6w9v62y8wr40d2r' },
+      mintedBlocks: expect.any(Number),
+      owner: { address: 'mwsZw8nF7pKxWH8eoKL9tPxTpaFkz7QeLU' },
+      operator: { address: 'mswsMVsyGMj1FzDMbbxw2QW3KvQAv2FKiy' },
       creation: { height: 0 },
       resign: {
-        tx: '0000000000000000000000000000000000000000000000000000000000000000',
+        tx: '',
```
fuxingloh, line 49:
```suggestion
        tx: undefined,
```

How about we make it optional, make it less misleading

src/module.api/masternode.controller.ts
```diff
@@ -43,36 +43,37 @@ export class MasternodeController {
...
+        address: info.ownerAddress
+      },
+      operator: {
+        address: info.operatorAddress
+      },
+      creation: {
+        height: info.creationHeight
+      },
+      resign: {
+        tx: info.resignTx ?? '',
```
fuxingloh, line 73:
```suggestion
        tx: info.resignTx,
```

fuxingloh (changes requested):

src/module.indexer/model/dftx/create.masternode.ts
fuxingloh: Don't really understand what you mean. But I guess we can leave this as it is.

ivan-zynesis (commented):
Looks good for DeFi Scan use.

Should we add extra indexing with owner address?

monstrobishi:
One moment gentleman, updating tests to reflect optional resign object

monstrobishi:
@ivan-zynesis I would like to see that in another PR

packages/whale-api-client/src/api/masternodes.ts
```diff
@@ -46,10 +50,11 @@ export interface MasternodeData {
   creation: {
     height: number
   }
-  resign: {
-    tx: string
+  resign?: {
+    tx?: string
```
fuxingloh, line 54:
```suggestion
    tx:: string
```

fuxingloh: since now resign is optional, tx don't need to be

fuxingloh (changes requested):

ivan-zynesis (approved):

fuxingloh (approved):

PR #349, opened 2021-08-19
masternode own indexer & model

canonbrother:

/kind feature

Listing particular user own masternode
eg: wallet calls `/{user collateral addres}/masternodes` to retrieve own mn
resp: 
```ts
[
      {
        id: 'mnA3',
        ownerAddress: 'mnA',
        operatorAddress: 'mnAc',
        creationHeight: 0,
        resignHeight: 0,
        mintedBlocks: 1,
        timelock: 0,
        local: true,
        block: { hash: '', height: 3, time: 0, medianTime: 0 }
      },
      {
        id: 'mnA2',
        ownerAddress: 'mnA',
        operatorAddress: 'mnAb',
        creationHeight: 0,
        resignHeight: 0,
        mintedBlocks: 1,
        timelock: 0,
        local: true,
        block: { hash: '', height: 2, time: 0, medianTime: 0 }
      },
      {
        id: 'mnA1',
        ownerAddress: 'mnA',
        operatorAddress: 'mnAa',
        creationHeight: 0,
        resignHeight: 0,
        mintedBlocks: 1,
        timelock: 0,
        local: true,
        block: { hash: '', height: 1, time: 0, medianTime: 0 }
      }
]
```

Fixes #feature

#### Additional comments?:
Blocked by 
https://github.com/DeFiCh/jellyfish/pull/534
https://github.com/DeFiCh/whale/pull/344
mn targetMultiplier

### DIFF
--- a/src/module.model/masternode.own.spec.ts
+++ b/src/module.model/masternode.own.spec.ts
@@ -1,0 +1,83 @@
+import { Database } from '@src/module.database/database'
+import { Test } from '@nestjs/testing'
+import { MemoryDatabaseModule } from '@src/module.database/provider.memory/module'
+import { LevelDatabase } from '@src/module.database/provider.level/level.database'
+import { MasternodeOwnMapper } from '@src/module.model/masternode.own'
+
+let database: Database
+let mapper: MasternodeOwnMapper
+
+beforeAll(async () => {
+  const app = await Test.createTestingModule({
+    imports: [MemoryDatabaseModule],
+    providers: [MasternodeOwnMapper]
+  }).compile()
+
+  database = app.get<Database>(Database)
+  mapper = app.get<MasternodeOwnMapper>(MasternodeOwnMapper)
+})
+
+beforeEach(async () => {
+  async function put (
+    masternodeId: string, ownerAddress: string, operatorAddress: string, height: number
+  ): Promise<void> {
+    await mapper.put({
+      id: masternodeId,
+      ownerAddress: ownerAddress,
+      operatorAddress: operatorAddress,
+      creationHeight: 0,
+      resignHeight: 0,
+      resignTx: undefined,
+      mintedBlocks: 1,
+      timelock: 0,
+      local: true,
+      block: { hash: '', height: height, time: 0, medianTime: 0 }
+    })
+  }
+
+  await put('mnA1', 'mnA', 'mnAa', 1)
+  await put('mnA2', 'mnA', 'mnAb', 2)
+  await put('mnA3', 'mnA', 'mnAc', 3)
+
+  await put('mnB1', 'mnB', 'mnBa', 20)
+})
+
+afterEach(async () => {
+  await (database as LevelDatabase).clear()
+})
+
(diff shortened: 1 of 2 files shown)


monstrobishi:
I'm already working on this: https://github.com/DeFiCh/whale/pull/344

monstrobishi:
@canonbrother

monstrobishi:
See here the stats indexed: https://github.com/DeFiCh/whale/pull/344/files#diff-291e2cc437b58d891c6210ee8df599508d50fcceceb32e3f0161fe37e39f8785R44

monstrobishi:
Sorry I just understood the purpose here, you want to look up MN info based on ownerAddress right. I think you should still build on top of my existing PR though as there must be some overlap

canonbrother:
> Sorry I just understood the purpose here, you want to look up MN info based on ownerAddress right. I think you should still build on top of my existing PR though as there must be some overlap

yea after i checked the PR #344, that is not for mine :sweat_smile: 
gonna separate another own masternode index

monstrobishi:
I suggest having a separate mapper but adding the remaining logic into create masternode indexer, most of the setup is there so you just need to call put on the mapper. Additionally you can reuse the model definition.

fuxingloh:
You can only have one masternode per address right? Since the UTXO is locked up.

We can pass in the `xpubkey`, but let's not do that (yet).

fuxingloh:
> You can only have one masternode per address right? Since the UTXO is locked up.
> 
> We can pass in the `xpubkey`, but let's not do that (yet).

Then we might not need this if we index the `owner` and operator in the masternode.

monstrobishi:
@fuxingloh I agree but I think @canonbrother point js that we need to look up by owner, so perhaps just another index or partition key needs to be added to the existing mn model?

fuxingloh:
> @fuxingloh I agree but I think @canonbrother point is that we need to look up by owner, so perhaps just another index or partition key needs to be added to the existing mn model?

Yeah, adding another index

fuxingloh:
Maybe something like this?

- `/masternodes/:id` existing
- `/address/:owner/masternodes/owner` 404 or 200 via owner index with key partition
- `/address/:operator/masternodes/operator` return [...] 200 via operator index with key+sort partition

canonbrother:
> I suggest having a separate mapper but adding the remaining logic into create masternode indexer, most of the setup is there so you just need to call put on the mapper. Additionally you can reuse the model definition.

yea its separated, will update logic on your indexer

> You can only have one masternode per address right? Since the UTXO is locked up.

user can have many opeAddr while user set up his mn like this
```
colAddr = wallet.get(0).getAddress()
createmn(colAddr, opeAddr1) // 1x collateral
createmn(colAddr, opeAddr2) // 2x collateral
createmn(colAddr, opeAddr3) // 3x collateral
```
the resp will be similar above

monstrobishi:
Is this on hold?

canonbrother:
> Is this on hold?

yea this one still on hold.. probably after `loan/vault` index

PR #355, opened 2021-08-23
/stats emission

### LINKED ISSUE #308: `/stats` - emission

Provide additional statistical data in `/stats` endpoint.

```json
{
  "emission": {
    "total": number,
    "masternode": number,
    "dex": number,
    "community": number,
    "anchor": number,
    "burned": number
  }
}
```

Total to include burned to keep consistency with other stats in the `/stats` endpoint.

/triage accepted
/area module-api
/priority important-soon

siradji:

/kind feature

Add emission to /stat endpoint

Fixes #308


siradji:
@monstrobishi  need your advice on this.
https://github.com/DeFiCh/whale/blob/7463d7c4543dc0dd106f3b86d7f9d9f2a380cd5d/src/module.api/stats.controller.ts#L92-L132

monstrobishi:
@siradji 

For `eunosHeight` you should be able to get this from `getBlockchainInfo`, it's underneath the soft forks object (although it's technically a hard fork)

Remember to cache the RPC requests.

Remember that the emissions reduction period, etc could be different depending on the network. See the relevant values in chainparams.cpp: 

Mainnet: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L112
Testnet: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L330
Regtest: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L682

For the regtest value perhaps go with isJellyfish by default, as this is likely the one we are using for all of our test cases.

For the numbers involving DFI amounts switch to BigNumber, i.e.:
```
reductionAmount <= 0.0000000001
```
Could be
```
reductionAmount.lte(new BigNumber(0.00001))
```

We might also need 5 decimal 0 there as I changed above because the actual cpp code divides an integer by 10000 and checks if it's 0 (See: https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L1143) 
```
            CAmount reductionAmount;
            for (size_t i = reductions; i > 0; --i)
            {
                reductionAmount = (nSubsidy * consensusParams.emissionReductionAmount) / 100000;
                if (!reductionAmount) {
                    nSubsidy = 0;
                    break;
                }

                nSubsidy -= reductionAmount;
            }
```

Do check  https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L1143 and other relevant sections in the CPP.

Overall the logic looks good, do also expand the 33.33 * 0.01 to 0.3333 etc, for example, use BigNumbers where you can.

monstrobishi:
I say go with a good implementation first and use the tests to check it's validity, we should be able to verify the reduction is correct by testing against the first vOut of every block, once the reduction has gone through (In mainnet that's 2 weeks, i.e. 2880*14 blocks), where as in jellyfish regtest it's `32690` so much longer.

You might want to test against a regtest container with `isJellyFish` off actually as the reduction period is too long for practical testing (And use the non-jellyfish value in your implementaion).

siradji:
> @siradji 
> 
> 
> 
> For `eunosHeight` you should be able to get this from `getBlockchainInfo`, it's underneath the soft forks object (although it's technically a hard fork)
> 
> 
> 
> Remember to cache the RPC requests.
> 
> 
> 
> Remember that the emissions reduction period, etc could be different depending on the network. See the relevant values in chainparams.cpp: 
> 
> 
> 
> Mainnet: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L112
> 
> Testnet: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L330
> 
> Regtest: https://github.com/DeFiCh/ain/blob/master/src/chainparams.cpp#L682
> 
> 
> 
> For the regtest value perhaps go with isJellyfish by default, as this is likely the one we are using for all of our test cases.
> 
> 
> 
> For the numbers involving DFI amounts switch to BigNumber, i.e.:
> 
> ```
> 
> reductionAmount <= 0.0000000001
> 
> ```
> 
> Could be
> 
> ```
> 
> reductionAmount.lte(new BigNumber(0.00001))
> 
> ```
> 
> 
> 
> We might also need 5 decimal 0 there as I changed above because the actual cpp code divides an integer by 10000 and checks if it's 0 (See: https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L1143) 
> 
> ```
> 
>             CAmount reductionAmount;
> 
>             for (size_t i = reductions; i > 0; --i)
> 
>             {
> 
>                 reductionAmount = (nSubsidy * consensusParams.emissionReductionAmount) / 100000;
> 
>                 if (!reductionAmount) {
> 
>                     nSubsidy = 0;
> 
>                     break;
> 
>                 }
> 
> 
> 
>                 nSubsidy -= reductionAmount;
> 
>             }
> 
> ```
> 
> 
> 
> Do check  https://github.com/DeFiCh/ain/blob/master/src/validation.cpp#L1143 and other relevant sections in the CPP.
> 
> 
> 
> Overall the logic looks good, do also expand the 33.33 * 0.01 to 0.3333 etc, for example, use BigNumbers where you can.

Okay got it. Thanks!

siradji:
@monstrobishi , please review this again

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,73 @@ export class StatsController {
...
+    if (burnInfo != null) {
+      burned = burnInfo.amount.plus(burnInfo.emissionburn).plus(burnInfo.feeburn).toNumber()
+    }
+
+    if (isMain) {
+      const emissionReductionPeriod = 32690 // Two weeks
+      const emissionReductionAmount = 1658
+      const reductions = Math.floor((nBlocks - eunosHeight) / emissionReductionPeriod)
+
+      let reductionAmount = new BigNumber(0)
```
jingyi2811, line 145: I don't think this line is necessary.

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,73 @@ export class StatsController {
...
+      for (let i = reductions; i > 0; i--) {
+        reductionAmount = new BigNumber(blockSubsidy * emissionReductionAmount / 100000)
```
jingyi2811, line 147:
```suggestion
        let eductionAmount = new BigNumber(blockSubsidy * emissionReductionAmount / 100000)
```

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,73 @@ export class StatsController {
...
+        if (reductionAmount.lte(0.00001)) {
+          blockSubsidy = 0
+          break
+        }
+
+        blockSubsidy -= reductionAmount.toNumber()
```
jingyi2811, line 153: Is that better if you do the `toNumber` conversion outside the loop?

monstrobishi:
Would like to see some tests. If you need some guidance you can follow some the tests here:

https://github.com/DeFiCh/ain/blob/epic/loans/src/test/dip1fork_tests.cpp#L158

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,73 @@ export class StatsController {
...
+      eunosHeight = blockInfo.softforks.eunos.height
+      nBlocks = blockInfo.blocks
+    }
+    if (burnInfo != null) {
+      burned = burnInfo.amount.plus(burnInfo.emissionburn).plus(burnInfo.feeburn).toNumber()
+    }
+
+    if (isMain) {
+      const emissionReductionPeriod = 32690 // Two weeks
+      const emissionReductionAmount = 1658
```
monstrobishi, line 142: I think use a BigNumber here, and remove the / 100000 (see below)

monstrobishi: And change to `1.658%`

src/module.api/stats.controller.ts
monstrobishi, line 147: Just use the proper `1.658%` (I guess that's 0.01658) then remove the division by 100000 here

monstrobishi:
This seems like it's only calculating post-eunos height emissions, see the C++ logic also checks for the height and returns the old subsidy values if it's pre eunos height:

https://github.com/DeFiCh/ain/blob/20392165c633470fd5b4922ea075f4fb6bf5c23e/src/validation.cpp

src/module.api/stats.controller.ts
siradji: Yep. Removed

packages/whale-api-client/__tests__/api/stats.test.ts
```diff
@@ -1,67 +1,112 @@
...
+    emissionReductionPeriod: 32690,
+    eunosHeight: 10000000
+  }
 
-afterAll(async () => {
-  try {
-    await service.stop()
-  } finally {
-    await container.stop()
+  function calculateCoinbaseRewards (blockReward: number, percentage: number): number {
```
monstrobishi, line 82: Any reason these functions are inside the test itself, and not in the source code?

monstrobishi: Oh you are using to compare against returned values. Are you able to hardcode and test against a range of values like the cpp test I linked?

siradji: Yes, it works. I just feel testing the return values from the endpoint make more sense. Any reason why we should keep it hardcoded? just curious.

monstrobishi: Because having the same logic in two places isn't testing against anything. It's just running the same function twice with the same parameters. The hardcoded parameters from the CPP test at the bear minimum provide us a baseline to ensure our logic produces the same output as the CPP version with all of the integer precision calculations.

monstrobishi: Thanks resolving this convo

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,75 @@ export class StatsController {
       })
     }
   }
+
+  private async getEmission (): Promise<StatsData['emission']> {
+    const burnInfo = await this.getBurnInfo()
+    const blockInfo = await this.getBlockChainInfo()
+
+    let eunosHeight = 0
```
monstrobishi, line 126:
```suggestion
    let eunosHeight = blockInfo?.softforks?.eunos?.height ?? 0
```

src/module.api/stats.controller.ts
```diff
@@ -37,11 +39,12 @@ export class StatsController {
       price: await this.cachedGet('price', this.getPrice.bind(this), 300),
       masternodes: {
         locked: masternodes.locked
-      }
+      },
+      emission: await this.cachedGet('emission', this.getEmission.bind(this), 1000)
```
monstrobishi, line 43: Maybe at a later stage `getEmission` can take a height instead, then at this level we pass the eunos height and burn info (caching those), but I'll approve for now

monstrobishi (dismissed):

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,75 @@ export class StatsController {
...
+  if (nHeight >= eunosHeight) {
+    for (let i = reductions; i > 0; i--) {
+      const reductionAmount = emissionReductionAmount.times(blockSubsidy)
+      if (reductionAmount.lte(0.00001)) {
+        blockSubsidy = 0
+        break
+      }
+      blockSubsidy -= reductionAmount.toNumber()
+    }
+    return blockSubsidy
```
jingyi2811, line 187: I think this line can be removed.

packages/whale-api-client/__tests__/api/stats.test.ts
```diff
@@ -3,66 +3,79 @@ import { StubWhaleApiClient } from '../stub.client'
...
+    return (blockReward * percentage) / 10000
   }
-})
 
-describe('stats', () => {
-  it('should get', async () => {
+  function calculateReductionHeight (reductions: number): number {
+    return consensusParams.eunosHeight + (reductions * consensusParams.emissionReductionPeriod)
```
jingyi2811, line 75:
```suggestion
    return consensusParams.eunosHeight + reductions * consensusParams.emissionReductionPeriod
```

src/module.api/stats.controller.ts
siradji: Yep. True ture

src/module.api/stats.controller.ts
siradji: Okay. I will open an issue for improvement after his gets merged.  Thanks for the assistance so far

siradji:
@monstrobishi , re-approval please.

monstrobishi (dismissed):

jingyi2811 (approved):

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,71 @@ export class StatsController {
...
+  }
+
+  private async getBurnInfo (): Promise<BurnInfo | undefined> {
+    return await this.cache.get<BurnInfo>('BURN_INFO', async () => {
+      return await this.rpcClient.account.getBurnInfo()
+    })
+  }
+}
+
+export function getBlockSubsidy (eunosHeight: number, nHeight: number = 0): number {
```
fuxingloh, line 169: Can this be implemented without loops?

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,71 @@ export class StatsController {
...
+    const communityRewards: Record<string, BigNumber> = {
+      masternode: new BigNumber(0.3333),
+      dex: new BigNumber(0.2445),
+      community: new BigNumber(0.0491),
+      anchor: new BigNumber(0.0002)
+    }
+
+    for (const reward in communityRewards) {
+      communityRewards[reward] = communityRewards[reward].times(blockSubsidy)
+    }
```
fuxingloh, line 144: reduce complexity, remove the loops

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,71 @@ export class StatsController {
...
+
+  private async getEmission (): Promise<StatsData['emission']> {
+    const burnInfo = await this.getBurnInfo()
+    const blockInfo = await this.getBlockChainInfo()
+
+    const eunosHeight = blockInfo?.softforks.eunos.height ?? 0
+    let burned = 0
+
+    if (burnInfo != null) {
+      burned = burnInfo.amount.plus(burnInfo.emissionburn).plus(burnInfo.feeburn).toNumber()
```
fuxingloh, line 130: won't be the burn be `1.0 - given rewards`, removing the need for getBurnInfo()

fuxingloh (changes requested):

src/module.api/stats.controller.ts
siradji: I do not think so as this is exactly how it is in the c++ side. @monstrobishi ??

monstrobishi: ^ yes this is an exact replica of C++ implementation

src/module.api/stats.controller.ts
monstrobishi: yes this loop can be unrolled, its just looping over all the types

packages/whale-api-client/__tests__/api/stats.test.ts
```diff
@@ -3,66 +3,79 @@ import { StubWhaleApiClient } from '../stub.client'
 import { StubService } from '../stub.service'
 import { WhaleApiClient } from '../../src'
 import { addPoolLiquidity, createPoolPair, createToken, getNewAddress, mintTokens } from '@defichain/testing'
+import { getBlockSubsidy } from '@src/module.api/stats.controller'
```
fuxingloh, line 6:
```suggestion
```

don't import controller logic for e2e tests, keep your testing code static as the numbers should always be the same at the same block height

src/module.api/stats.controller.ts
fuxingloh: @siradji is this true?

src/module.api/stats.controller.ts
```diff
@@ -115,6 +118,67 @@ export class StatsController {
...
+    }
+
+    return {
+      total: blockSubsidy,
+      anchor: communityRewards.anchor.toNumber(),
+      dex: communityRewards.dex.toNumber(),
+      community: communityRewards.community.toNumber(),
+      masternode: communityRewards.masternode.toNumber(),
+      burned
+    }
```
fuxingloh, line 149: unnecessary redeclaration, you can compress this

src/module.api/stats.controller.ts
fuxingloh: is there a way to write this without loops? (regardless of how C++ implement it)

fuxingloh (changes requested):

src/module.api/stats.controller.ts
siradji: no i don't think so 1.0 - anything will be a negative.

src/module.api/stats.controller.ts
siradji: Let me study the code more and understand it better. Perhaps it is possible

siradji:
https://github.com/DeFiCh/whale/pull/355#discussion_r716202486

@fuxingloh it is true. The burned total is the amount  + feeburn. 
https://github.com/DeFiCh/whale/blob/9698a05ebecdb86c74f6521c6d14b1017a0d56ad/src/module.api/stats.controller.ts#L83-L91

siradji:
https://github.com/DeFiCh/whale/pull/355#discussion_r716202914

@fuxingloh , i checked and to my understanding i can find an alternative. perhaps @monstrobishi you can put your input here.

siradji:
@fuxingloh , i need to get this merged before i can continue working on the supply data.

fuxingloh (approved):

PR #367, opened 2021-08-27
`script.activity.v2.indexer` - interpret DfTx and index token activity by script

ivan-zynesis:

/kind feature

For each wallet (script) to list all token activity history

Fixes #


src/module.indexer/model/token_activity/utxos.to.account.ts
```diff
@@ -0,0 +1,28 @@
...
+    for (const { script, balances } of scriptBalances) {
+      const scriptPubKey = script.stack[1] as OP_PUSHDATA
+      for (const bal of balances) {
+        result.push({
+          script: {
+            type: scriptPubKey.type,
+            hex: scriptPubKey.hex
+          },
+          type: 'utxos-to-account-gain',
+          tokenId: bal.token,
```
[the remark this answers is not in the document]
fuxingloh: Yeah, it's different things totally that might change in the future. Probably shouldn't mix

src/module.indexer/model/token_activity/_abstract.ts
```diff
@@ -0,0 +1,81 @@
...
+    private readonly mapper: ScriptActivityMapper
+  ) {
+  }
+
+  async index (block: RawBlock, txns: Array<DfTxTransaction<T>>): Promise<void> {
+    let i = 0
+    for (const txn of txns) {
+      const activities = await this.extractTokenActivities(txn.dftx.data)
+      for (const activity of activities) {
+        const simpleScriptActivity = this.mapToGenericScriptActivity(block, txn, activity, i++)
```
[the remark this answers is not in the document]
fuxingloh: I think we should design a "ScriptActivityV2". Hard to achieve compatibility as we can't deploy this into the existing endpoint as it would break compatibility.

The UTXO part is not well written to support DfTX in mind causing incompatible issues.

I would think a v2 would have a simplified interface for mixing data such as this:
```ts
{
  script: ...,
  utxo: {},
  dftx: {}
}
```

src/module.model/script.activity.v2.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  async put (activity: ScriptActivityV2): Promise<void> {
+    return await this.database.put(ScriptActivityV2Mapping, activity)
+  }
+
+  async delete (id: string): Promise<void> {
+    return await this.database.delete(ScriptActivityV2Mapping, id)
+  }
+
+  async insert (activity: ScriptActivityV2): Promise<void> {
```
ivan-zynesis, line 45: Thinking adding this be 100% certain no data loss if it is not too much of impact.

src/module.model/script.activity.v2.ts
```diff
@@ -0,0 +1,89 @@
...
+  value: string // -------------| output value stored as string, string as decimal: 0.0000
+  tokenId: number // -----------| 0: DFI, 1: BTC, etc
+  type: ScriptActivityV2Type
+
+  category: ScriptActivityCategory
+
+  utxo?: {
+    txid: string
+    n: number
+  }
```
ivan-zynesis, line 88: Just realized `dftx?: {}` has no content, not that I aware off so far.

src/module.model/script.activity.v2.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  script: {
+    type: string
+    hex: string
+  }
+
+  value: string // -------------| output value stored as string, string as decimal: 0.0000
+  tokenId: number // -----------| 0: DFI, 1: BTC, etc
+  type: ScriptActivityV2Type
```
fuxingloh, line 81: I think these can be moved to UTXO

src/module.model/script.activity.v2.ts
fuxingloh: Should we design something for it or just hold the RawDfTx?

ivan-zynesis: Make sense, client will be able to reconstruct CDfTx.

src/module.indexer/model/script.activity.v2/common.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @param block
+ * @param txid
+ * @param n an unique serial number for each script.activity.v2 within each block/txid/category
+ * @returns {string}
+ */
+export function mapId (block: RawBlock, txid: string, category: Category, n: number): string {
+  const height = HexEncoder.encodeHeight(block.height)
+  const index = HexEncoder.encodeHeight(n) // to convert single number into 4 bytes long
+  return `${height}${txid}${categoryHexMap[category]}${index}`
```
ivan-zynesis, line 26: I rearranged the order in id construction, thinking `txid` should have higher priority than `category` so the order will be closer to actual chronological order as how each transaction created.

ivan-zynesis:
DEX output are calculated by consensus real time. Info not directly available on chain thus not indexed here.

PR #370, opened 2021-08-30
Add `displaySymbol`

monstrobishi:

/kind chore

Add `displaySymbol`

Fixes #


monstrobishi:
Fixes #370 361

src/module.api/token.controller.e2e.ts
```diff
@@ -109,6 +112,7 @@ describe('list', () => {
       id: '3',
       symbol: 'DBTC-DET',
       symbolKey: 'DBTC-DET',
+      displaySymbol: 'dDBTC-DET',
```
fuxingloh, line 115:
```suggestion
      displaySymbol: 'dDBTC-dDET',
```

should it be this instead?

fuxingloh (changes requested):

src/module.api/token.controller.e2e.ts
monstrobishi: Yes I saw this. I decided it doesn't make sense to change at this stage.

See, the d prefix is already handled at the poolpair level by:
```
    tokenA: {
      displaySymbol: info.idTokenA === '0' ? symbolA : `d${symbolA}`,
    },
    tokenB: {
      displaySymbol: info.idTokenB === '0' ? symbolB : `d${symbolB}`,
    },
```

The above scenario is when you are viewing the LP token for a particular poolpair.

You could say then, why not add some special handling to tokens with `isLP`, and do `symbol.split('-').map(x => `d${x}`).join('-')`. You run into 2 problems:
- Technically, the LP token symbol is not restricted to `tokenA-tokenB`, infact in `createpoolpair` you can assign any symbol, that is just the default when you don't specify a param
- The token symbol is trimmed to 8 chars, so `ETHEREUM-DOGE` would result in just `ETHEREUM` with no `-`, breaking the above code

I think for consistency I can disable the `d` prefix for `isLP` tokens, that's probably the most sane solution right now.

monstrobishi: I've implemented my suggested change which is to disable the `d` prefix on isLP tokens (See above comment)

fuxingloh (approved):

PR #373, opened 2021-08-31
Index `create.token`

monstrobishi:

/kind feature

This indexes `create.token`. Initially I wanted to index `update.token` too (Hence the branch name), but this would require a secondary index as `update.token` uses the txid as a reference, so I decided that was out of scope.

Fixes #


src/module.model/token.ts
```diff
@@ -0,0 +1,90 @@
+import { Model, ModelMapping } from '@src/module.database/model'
+import { Injectable } from '@nestjs/common'
+import { Database, SortOrder } from '@src/module.database/database'
+import { HexEncoder } from './_hex.encoder'
+import BigNumber from 'bignumber.js'
+
+export const DCT_ID_START = 128
```
fuxingloh, line 7: Well, this number is way too low LOL

src/module.indexer/model/dftx/create.token.ts
```diff
@@ -0,0 +1,62 @@
...
+      await this.tokenMapper.delete(`${tokenId - 1}`)
+    }
+  }
+
+  async getNextTokenID (isDAT: boolean): Promise<number> {
+    const latest = isDAT ? await this.tokenMapper.getLatestDAT()
+      : await this.tokenMapper.getLatestDST()
+
+    if (latest === undefined) {
+      // Default to 1 as 0 is reserved for DFI
```
fuxingloh, line 51: we should index DFI token on indexGenesis method.

src/module.indexer/model/dftx/create.token.ts
```diff
@@ -0,0 +1,62 @@
...
+  async invalidate (_: RawBlock, txns: Array<DfTxTransaction<TokenCreate>>): Promise<void> {
+    for (const { dftx: { data } } of txns) {
+      const tokenId = await this.getNextTokenID(data.isDAT)
+      await this.tokenMapper.delete(`${tokenId - 1}`)
+    }
+  }
+
+  async getNextTokenID (isDAT: boolean): Promise<number> {
+    const latest = isDAT ? await this.tokenMapper.getLatestDAT()
+      : await this.tokenMapper.getLatestDST()
```
fuxingloh, line 48: you should separate your logic into `if (isDat) {} else {}` with 2 major code blocks, instead of conditionally running your logic in multiple code blocks. Easier to read and maintain long term.

fuxingloh (changes requested):

src/module.indexer/model/dftx/create.token.ts
monstrobishi: Oh yes I had this in monstrobishi/index-poolpairs, will cherry pick over

src/module.indexer/model/dftx/create.token.ts
```diff
@@ -0,0 +1,72 @@
...
+      await this.tokenMapper.delete(`${tokenId - 1}`)
+    }
+  }
+
+  async getNextTokenID (isDAT: boolean): Promise<number> {
+    if (isDAT) {
+      const latest = await this.tokenMapper.getLatestDAT()
+      if (latest === undefined) {
+        // Default to 1 as 0 is reserved for DFI
+        return 1
```
fuxingloh, line 51: throw error?

fuxingloh (changes requested):

src/module.indexer/model/dftx/create.token.ts
monstrobishi: Oh yes, since DFI is already indexed?

fuxingloh (approved):

PR #374, opened 2021-08-31
Fix block minted issue

monstrobishi:

/kind fix

This fixes an issue in the block minted mapper that causes the following error print out and causes the `indexGenesis` function to exit early:

```
[Nest] 10720  - 31/08/2021, 8:00:46 pm   ERROR [RPCBlockProvider] failed exceptionally
[Nest] 10720  - 31/08/2021, 8:00:46 pm   ERROR [RPCBlockProvider] ReadError: key cannot be `null` or `undefined`
```

Fixes #


monstrobishi:
Added test case, note that I had to make `indexGenesis` public for the test

fuxingloh (dismissed):

PR #385, opened 2021-09-09
`/blocks` & `/transactions` improvement

### LINKED ISSUE #383: `/blocks` & `/transactions` API improvement

While working on DeFi Scan, I noticed 3 inconsistencies:

1. `blocks/:id/transactions` is not ordered by the first appearance in the block, currently it's sorted via txid. We need it to be sorted by the first appearance.
2. `/blocks` API `Block` model should include block reward, by checking the first transaction vout in the block.
3. `/transactions` API `Transaction` model should include total UTXO value out, the summing of all `vout` value.

/priority urgent-now
/triage accepted
/area module-api module-model module-database
/assign @monstrobishi

monstrobishi:

/kind fix

1. `blocks/:id/transactions` is not ordered by the first appearance in the block, currently it's sorted via txid. We need it to be sorted by the first appearance.
2. `/blocks` API `Block` model should include block reward, by checking the first transaction vout in the block.
3. `/transactions` API `Transaction` model should include total UTXO value out, the summing of all `vout` value.

See #383 https://github.com/DeFiCh/whale/issues/383

Fixes #383

### DIFF
--- a/packages/whale-api-client/__tests__/api/blocks.test.ts
+++ b/packages/whale-api-client/__tests__/api/blocks.test.ts
@@ -53,7 +53,7 @@ describe('list', () => {
     const first = await client.blocks.list(35)
 
     expect(first.length).toStrictEqual(35)
-    expect(first[0]).toStrictEqual(ExpectedBlock)
+    expect(first[0]).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
     expect(first[0].height).toBeGreaterThanOrEqual(100 - 35)
     expect(first[34].height).toBeGreaterThanOrEqual(first[0].height - 35)
 
@@ -69,14 +69,14 @@ describe('list', () => {
     const first = await client.blocks.list(100)
 
     expect(first.length).toStrictEqual(60)
-    expect(first[0]).toStrictEqual(ExpectedBlock)
+    expect(first[0]).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
   })
 
   it('should get paginated list of blocks when next is out of range', async () => {
     const blocks = await client.blocks.list(15, '1000000')
     expect(blocks.length).toStrictEqual(15)
 
-    expect(blocks[0]).toStrictEqual(ExpectedBlock)
+    expect(blocks[0]).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
     expect(blocks[0].height).toBeGreaterThanOrEqual(40)
   })
 
@@ -90,7 +90,7 @@ describe('list', () => {
     const blocks = await client.blocks.list(60)
     expect(blocks.length).toBeGreaterThanOrEqual(40)
 
-    expect(blocks[0]).toStrictEqual(ExpectedBlock)
+    expect(blocks[0]).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
     expect(blocks[0].height).toBeGreaterThanOrEqual(40)
   })
 })
@@ -99,14 +99,14 @@ describe('get', () => {
   it('should get block through height', async () => {
     const block = await client.blocks.get('37')
 
-    expect(block).toStrictEqual(ExpectedBlock)
+    expect(block).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
     expect(block?.height).toStrictEqual(37)
   })
 
   it('should get block through hash', async () => {
     const blockHash = await container.call('getblockhash', [37])
     const block = await client.blocks.get(blockHash)
-    expect(block).toStrictEqual(ExpectedBlock)
+    expect(block).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
     expect(block?.height).toStrictEqual(37)
   })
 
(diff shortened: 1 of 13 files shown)


packages/whale-api-client/__tests__/api/blocks.test.ts
```diff
@@ -53,7 +53,7 @@ describe('list', () => {
     const first = await client.blocks.list(35)
 
     expect(first.length).toStrictEqual(35)
-    expect(first[0]).toStrictEqual(ExpectedBlock)
+    expect(first[0]).toStrictEqual({ ...ExpectedBlock, reward: '33.33000000' })
```
fuxingloh, line 56: I think we can add 33.33 into `ExpectedBlock`

packages/whale-api-client/__tests__/api/transaction.test.ts
```diff
@@ -83,7 +83,8 @@ describe('get', () => {
       weight: expect.any(Number),
       lockTime: expect.any(Number),
       vinCount: expect.any(Number),
-      voutCount: expect.any(Number)
+      voutCount: expect.any(Number),
+      totalVOut: expect.any(String)
```
fuxingloh, line 87:
```suggestion
      totalVoutValue: expect.any(String)
```

packages/whale-api-client/src/api/transactions.ts
```diff
@@ -54,6 +54,7 @@ export interface Transaction {
   lockTime: number
   vinCount: number
   voutCount: number
+  totalVOut: string
```
fuxingloh, line 57:
```suggestion
  totalVoutValue: string
```

src/module.model/transaction.ts
```diff
@@ -51,6 +51,7 @@ export class TransactionMapper {
  */
 export interface Transaction extends Model {
   id: string // ----------------| unique id of the transaction, same as the txid
+  sort: string // --------------| height-order
```
fuxingloh, line 54: do we need height, maybe order will do

src/module.indexer/model/transaction.ts
```diff
@@ -20,9 +25,10 @@ export class TransactionIndexer extends Indexer {
     }
   }
 
-  map (block: RawBlock, txn: defid.Transaction): Transaction {
+  map (block: RawBlock, txn: defid.Transaction, order: number, totalVOut: string): Transaction {
     return {
       id: txn.txid,
+      sort: `${block.hash}-${HexEncoder.encodeHeight(order)}`,
```
fuxingloh, line 31: don't think we should add `block.hash` since it's already sort of block.hash

fuxingloh (changes requested):

src/module.indexer/model/transaction.ts
monstrobishi: I think it's needed because block hash is the partition key. If I try to query without prepending block hash I get no output

src/module.api/transaction.controller.ts
```diff
@@ -50,3 +51,20 @@ export class TransactionController {
     })
   }
 }
+
+function mapTransaction (tx: ModelTransaction): Transaction {
```
fuxingloh, line 55: does this remove, `sort`? we should just include it

src/module.api/block.controller.ts
```diff
@@ -57,3 +68,26 @@ export class BlockController {
     })
   }
 }
+
+function mapBlock (block: ModelBlock): Block {
```
fuxingloh, line 72: I don't think we need this?

src/module.indexer/model/transaction.ts
fuxingloh: we don't need it because the sort is part of the block.hash partition key,

1. txn 0: [block.hash: 64len, order: 0]
2. txn 1: [block.hash: 64len, order: 1]

fuxingloh (changes requested):

src/module.api/block.controller.ts
monstrobishi: Ok I thought it would be cleaner to explicity map whale Block into whale api client Block, but I can remove this

monstrobishi: Ok I get my mistake now, the server side controller doesn't care or know about the API client structures. There are some existing instances where the controller returns API client structures though so we may want to go back and refactor those.

fuxingloh: Yeah, it's one of the ugly patterns we have now. Something to refactor when we migrate to jellyfish. Right now it's just nice we have a perfect match from client to server.

fuxingloh:
/kind feature because this is a breaking change

src/module.api/block.controller.ts
```diff
@@ -37,11 +37,21 @@ export class BlockController {
   async get (@Param('id') hashOrHeight: string): Promise<Block | undefined> {
     const height = parseHeight(hashOrHeight)
     if (height !== undefined) {
-      return await this.blockMapper.getByHeight(height)
+      const block = await this.blockMapper.getByHeight(height)
+      if (block === undefined) {
+        return undefined
+      }
+
+      return block
```
fuxingloh, line 45: I think you can revert these changes, as it's already returning undefined if undefined?

src/module.api/block.controller.ts
```diff
@@ -37,11 +37,21 @@ export class BlockController {
...
     }
 
     if (isSHA256Hash(hashOrHeight)) {
-      return await this.blockMapper.getByHash(hashOrHeight)
+      const block = await this.blockMapper.getByHash(hashOrHeight)
+      if (block === undefined) {
+        return undefined
+      }
+
+      return block
```
fuxingloh, line 54: same for this?

fuxingloh (changes requested):
minor change request

src/module.api/block.controller.ts
fuxingloh:
```suggestion
      return await this.blockMapper.getByHeight(height)
```

src/module.api/block.controller.ts
fuxingloh:
```suggestion
      return await this.blockMapper.getByHash(hashOrHeight)
```

fuxingloh (approved):

PR #389, opened 2021-09-15
`/anchors` endpoint

### LINKED ISSUE #322: Add listAnchors endpoint
`listAnchors` endpoint on `masternode` controller
#### Why is this needed:
Will be needed in `defi/Scan`

siradji:

/kind feature

Fixes #322


siradji:
Docker buildx and sanity test are failing and i think it is related the new package i installed.  How do i resolve this?

image

siradji:
@fuxingloh , I found out you want to add listanchors to masternode controller alone. I have already created a controller for anchors. Should i keep it or just go as planned?

siradji:
@fuxingloh could you give me some pointers on why my test is failing?

Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY LICENSE ./
 COPY package.json ./
 COPY package-lock.json ./
 
-RUN npm ci
```
fuxingloh, line 12: remove this

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor (
+    protected readonly rpcClient: JsonRpcClient
+  ) {
+  }
+
+  /**
+   *  List anchors
+   */
+  @Get('')
+  async list (): Promise<AnchorData[]> {
```
fuxingloh, line 17: missing pagination object

src/module.api/anchors.controller.e2e.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  app = await createTestingApp(container)
+  controller = app.get(AnchorsController)
+
+  await setup()
+})
+
+afterAll(async () => {
+  await stopTestingApp(container, app)
+  await tGroup.stop()
```
fuxingloh, line 27: `stopTestingApp` stops a container, tGroup also stops a container. It seems like you are attempting to stop an already stopped container.

try:
- app.stop
- group.stop

fuxingloh (changes requested):

src/module.api/anchors.controller.ts
siradji: @fuxingloh , listanchors rpc does not supports pagination. I believe there is way around this on the whale side. Could you point me to that?

Dockerfile
siradji: I changed it but got a build error. Another way to solve this is to commit my package-lock file.

packages/whale-api-client/src/api/anchors.ts
```diff
@@ -0,0 +1,36 @@
...
+ * DeFi whale endpoint for anchors related services.
+ */
+
+export class Anchors {
+  constructor (private readonly client: WhaleApiClient) {}
+
+  /**
+   * List anchors
+   */
+
```
canonbrother, line 13:
```suggestion
```

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   *  List anchors
+   */
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<AnchorData>> {
+    const result = await this.rpcClient.spv.listAnchors()
```
fuxingloh, line 22: Missing pagination implementation via `maxBtcHeight`

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    defiBlock: {
+      height: anchors.defiBlockHeight,
+      hash: anchors.defiBlockHash
+    },
+    previousAnchor: anchors.previousAnchor,
+    rewardAddress: anchors.rewardAddress,
+    confirmations: anchors.confirmations,
+    signatures: anchors.signatures,
+    active: anchors.active,
+    anchorCreationHeight: anchors.anchorCreationHeight
```
fuxingloh, line 47:
```suggestion
    id: id.toString(),
    btc: {
      block: {
        height: anchors.btcBlockHeight,
        hash: anchors.btcBlockHash,
      },
      txn: {
        hash: anchors.btcTxHash
      },
      confirmations: anchors.confirmations,
    },
    dfi: {
      block: {
        height: anchors.defiBlockHeight,
        hash: anchors.defiBlockHash
      }
    },
    active: anchors.active,
    signatures: anchors.signatures,
    creationHeight: anchors.anchorCreationHeight,
    rewardAddress: anchors.rewardAddress,
    previousAnchor: anchors.previousAnchor,
```

packages/whale-api-client/src/api/anchors.ts
```diff
@@ -0,0 +1,35 @@
...
+ * DeFi whale endpoint for anchors related services.
+ */
+
+export class Anchors {
+  constructor (private readonly client: WhaleApiClient) {}
+
+  /**
+   * List anchors
+   */
+  async list (): Promise<AnchorData[]> {
```
fuxingloh, line 13: pagination missing

fuxingloh (changes requested):

src/module.api/anchors.controller.ts
siradji: I have implemented the pagination but i think it will suffice for now. We need to change it and find a better way in the long way

siradji:
@fuxingloh, pagination has been implemented via minBtcHeight and maxBtcHeight. I think it is not the most ideal way and we need some better in the future.

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   *  List anchors
+   */
+
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<AnchorData>> {
+    const result = await this.rpcClient.spv.listAnchors({
+      maxBtcHeight: query.next !== undefined ? Number(query.next) + query.size : query.size,
```
fuxingloh, line 24: this won't really work because anchors have gaps, let's hold onto this PR for now. We need anchors RPC to support size pagination @surangap

siradji: Yeah that is what i have thought.

fuxingloh: Yeah, feel free to sound it upstream, if something is missing for your usecase.

surangap: @fuxingloh so the problem here is that since anchors have gaps, even though query for `query.size` records, it will not retrieve that amount. it will take the input as `maxBtcHeight`

The internal implementation of `listanchors` RPC is a bit different from other `list*` RPCs. especially from where the data is retrieved. it retrieves from a multi-index container, not from the db.

But we can implement the pagination with `startBTCHeight, number of records` for sure.
Ref: https://github.com/DeFiCh/ain/blob/2baae1d961a72640cc1950bbaecc9b9537dbd521/src/masternodes/anchors.h#L220

fuxingloh: Yup the problem with min/max height querying is you can't do size-based querying. If you attempt to do it like the above implementation, any large gaps will result in end of pagination. (min=next, max=next+size).

With start height + the number of records, is more deterministic as you can start searching from height with a max of n results. n[size] = the next "next token".

surangap: https://github.com/DeFiCh/ain/issues/910

siradji: @surangap, how are we getting on this?

fuxingloh:
Close in favor of #719

PR #392, opened 2021-09-21
add LP_DAILY_DFI_REWARD to stats api

p3root:
/kind feature

Add some more infos to the stats api


p3root:
Fix code duplication in the next PR

Issue #51, opened 2021-05-09
refactor `module.database`


fuxingloh:
#### What would you like to be added:

Similar to `module.indexer`, the current database implementation is done for the reasons of speed. We would need to refactor `module.database` implementation to be cleaner and faster.

- [ ] Setup new module for concurrent development
  - #98
- [ ] Remove `id` make it automated
- [ ] Return Keys for easier PagedResponse forming

fuxingloh:
Unassigned as it will be moved to `DeFiCh/jellyfish` in Q4.

Issue #50, opened 2021-05-09
refactor `module.indexer`


fuxingloh:
#### What would you like to be added:

Refactor `module.indexer` code to be cleaner. The current implementation is done in a jiffy which is necessitated for speed. We would need to refactor `module.indexer` with a cleaner design for better evolution of the code. Might be worth while to implement CQRS design pattern for indexing.

- [ ] Automated Invalidating via Prune, Undo, Redo e.t.c
- [ ] Separation of DfTx for transport and storage.
- [ ] DfTx Abstraction

fuxingloh:
`module.indexer` should also be refactored into an isolated single purpose module/package where all it does is indexing.

monstrobishi:
@fuxingloh Are indexers by design append-only? Should we enforce that as a rule? It would make automated invalidation much simpler

fuxingloh:
> @fuxingloh Are indexers by design append-only? Should we enforce that as a rule? It would make automated invalidation much simpler

Initially thinking it from a graph theory perspective, it makes sense but the more complex implementation e.g. `aggregation` pattern doesn't really follow the traditional ledger approach as it's aggregating latest balances when there is balance/movement activity

fuxingloh:
Unassigned as it will be moved to `DeFiCh/jellyfish` in Q4.

PR #407, opened 2021-09-27
add addresses to vout

p3root:
/kind feature
Add Addresses to vout information

### DIFF
--- a/src/module.indexer/model/_vout_finder.ts
+++ b/src/module.indexer/model/_vout_finder.ts
@@ -20,7 +20,8 @@ export class VoutFinder {
         tokenId: vout.tokenId,
         script: {
           type: vout.scriptPubKey.type,
-          hex: vout.scriptPubKey.hex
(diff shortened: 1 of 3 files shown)


fuxingloh:
Hey, thanks for opening #407 and #393. I will review, add tests and merge this in soon. 

Just to let you know the direction of this project, we are building 2 APIs concurrently. as part of https://github.com/DeFiCh/jellyfish/issues/580 and Rust node move. We are separating and moving more indexing responsibilities to jellyfish. 

The current API (this project) will be a full node compatible build and will still be worked on.

We are also working on another API that doesn't rely on defid, this new API which is in planning now will resolve the mountain of issues we face while building this API namely https://github.com/DeFiCh/whale/issues/50 and https://github.com/DeFiCh/whale/issues/51. This move is necessary to resolve the technical debt we accumulated in the early days that bring out Ocean API fast.

PR #432, opened 2021-10-04
active/next price controller level

monstrobishi:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:

active/next price indexing

#### Which issue(s) does this PR fixes?:

Fixes #

PR #433, opened 2021-10-04
`OraclePriceActive` `ActivePriceIndexer`

monstrobishi:

/kind feature

active/next price indexer level

Fixes #


src/module.indexer/model/dftx/set.oracle.data.ts
```diff
@@ -98,10 +98,33 @@ export class SetOracleDataIndexer extends DfTxIndexer<SetOracleData> {
...
     }
 
+    const aggregatedAmount = aggregated.total.dividedBy(aggregated.weightage).toFixed(8)
+    const key = `${token}-${currency}`
+    const previous = await this.aggregatedMapper.query(key, 1)
+
+    let activePrice = '0.00000000'
+    let nextPrice = '0.00000000'
+
+    if (block.height % 120 === 0) {
```
fuxingloh, line 108: It's refreshing every 120 blocks? It's not taking the 120 blocks prior to the current block?

monstrobishi: yes this is an attempt to mimic the C++ side implementation: 

https://github.com/DeFiCh/ain/pull/750/files#diff-97c3a52bc5fad452d82670a7fd291800bae20c7bc35bb82686c2c0a4ea7b5b98R3103

Every 120 blocks, active price becomes next price, and next price becomes the aggregated price

monstrobishi: this WIP I will move 120 into a constant later

fuxingloh: It's not about it being a constant.

This is because `block.height % 120 === 0` rule at `SetOracleData`, will only trigger every time there is a price with block%120 that come in the retrigger the next price update. If there isn't a DfTx for that block it will create gaps within the intervals.

fuxingloh: I believe C++ implementation is more on the RPC side? Hence it didn't matter?

fuxingloh: Easier if we use #432, but I favor the indexer level implementation because it allows history reading of active prices.

monstrobishi: Interesting, I was originally under the impression that OraclePriceAggregatedMapper was triggering on every block regardless, but looking at `mapPriceFeeds` I can see it only triggers when a relevant tx comes in.

The C++ implementation is not RPC side, you can see on the PR I linked above.

I got your answer that you favour the Indexer implementation so I will proceed with that. I will move the code so that it sits somewhere that is triggered every block rather than only when a `SetOracleData` comes in.

monstrobishi: Still just brainstorming here, I've implemented a slightly different approach, do you feel this works better? I can go ahead with tests/implementation details if this looks ok

fuxingloh: Also brainstorming here:

Slightly conflicted with the introduction of a new Model, I am leaning towards where we can have prices data only if there are prices activity.

monstrobishi: I'm open to suggestions, I think for now in order to accurately get active/next price behaviour we need to roll over every 120 blocks. Sure I can avoid writing to db if active/next price is the same, but the unique frequency of updates warrants a new model IMO.

src/module.indexer/model/active.price.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor (
+    private readonly aggregatedMapper: OraclePriceAggregatedMapper,
+    private readonly activePriceMapper: OraclePriceActiveMapper,
+    private readonly priceTickerMapper: PriceTickerMapper
+  ) {
+    super()
+  }
+
+  async index (block: RawBlock): Promise<void> {
+    if (block.height % BLOCK_INTERVAL !== 0) {
```
ivan-zynesis, line 25: Has cpp changed this? I remember previously %BLOCK_INTERVAL just has to be a constant. And each ticker may have different start point.

ivan-zynesis, line 25: Has cpp changed this? I remember previously %BLOCK_INTERVAL just has to be a constant. And each ticker may have different start point.

src/module.indexer/model/active.price.ts
```diff
@@ -0,0 +1,86 @@
...
+    const activePrice = previous?.next !== undefined ? previous.next : previous?.active
+
+    return {
+      id: `${ticker.id}-${block.height}`,
+      key: ticker.id,
+      valid: this.isLive(activePrice, nextPrice),
+      block: { hash: block.hash, height: block.height, medianTime: block.mediantime, time: block.time },
+      active: activePrice,
+      next: nextPrice,
+      sort: HexEncoder.encodeHeight(block.mediantime) + HexEncoder.encodeHeight(block.height)
```
ivan-zynesis, line 53: Block time is complementary of height, I think can remove the medianTime here

src/module.indexer/model/active.price.ts
```diff
@@ -0,0 +1,86 @@
...
+const DEVIATION_THRESHOLD = 0.3
+const BLOCK_INTERVAL = 120
+const MINIMUM_LIVE_ORACLES = 2
+
+@Injectable()
+export class ActivePriceIndexer extends Indexer {
+  constructor (
+    private readonly aggregatedMapper: OraclePriceAggregatedMapper,
+    private readonly activePriceMapper: OraclePriceActiveMapper,
+    private readonly priceTickerMapper: PriceTickerMapper
```
canonbrother, line 19: is that good to have network differentiate the `BLOCK_INTERVAL` ??  
120 mainnet | 6 regtest
:thinking: 
```suggestion
    private readonly priceTickerMapper: PriceTickerMapper
    @Inject('NETWORK') protected readonly network: NetworkName
```

canonbrother (commented):
overall good for me though

src/module.indexer/model/active.price.ts
monstrobishi: any other suggestions?

monstrobishi: Oh right I got you, yes will integrate this

src/module.indexer/model/active.price.ts
monstrobishi: Yes block height is from 0, that was changed once and then changed back again

monstrobishi:
@canonbrother @ivan-zynesis now that we upgraded jellyfish to 2.1.0 (This PR was submitted a while ago), I'll make a note to add some tests to compare with RPC results to be more confident

src/module.indexer/model/active.price.ts
monstrobishi: I'm not 100% certain there some blocks can still appear out of order

fuxingloh: Blocks can appear out of order and by indexing with `median time` your ActivePrice will be out of order too

packages/whale-api-client/src/api/prices.ts
```diff
@@ -154,3 +159,34 @@ export interface PriceOracle {
...
+
+  active?: {
+    amount: string
+    weightage: number
+    oracles: {
+      active: number
+      total: number
+    }
+  }
+  next: {
```
fuxingloh, line 176:
```suggestion
  next?: {
```

src/module.model/oracle.price.active.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  async delete (id: string): Promise<void> {
+    return await this.database.delete(OraclePriceActiveMapping, id)
+  }
+}
+
+export interface OraclePriceActive extends Model {
+  id: string // ---------| token-currency-height
+  key: string // --------| token-currency
+  sort: string // -------| medianTime-height
```
fuxingloh, line 48:
```suggestion
  sort: string // -------| height
```

src/module.indexer/model/active.price.ts
fuxingloh:
```suggestion
      sort: HexEncoder.encodeHeight(block.height)
```

src/module.indexer/model/active.price.ts
```diff
@@ -0,0 +1,93 @@
...
+    const activePrice = new BigNumber(active.amount)
+    const nextPrice = new BigNumber(next.amount)
+
+    return activePrice.gt(0) &&
+            nextPrice.gt(0) &&
+            nextPrice.minus(activePrice).abs().lt(activePrice.times(DEVIATION_THRESHOLD))
+  }
+
+  private isAggregateValid (aggregate: OraclePriceActive['next']): boolean {
+    return aggregate?.oracles !== undefined &&
```
fuxingloh, line 78: Checking that the oracle has recently published, the last aggregate can be taken from a long time ago since this is the latest data feed.

fuxingloh (changes requested):

src/module.indexer/model/active.price.ts
fuxingloh: Was thinking about this further, since this entity is only created in every `120` block interval, the "appear out of order" is pretty irrelevant in this context. I guess using height is just for consistency.

PR #440, opened 2021-10-05
simplify stats emission logic and testing

fuxingloh:
no point testing duplicated code logic

/kind fix

Code too complex, testing is redundant, no point duplicating code just to test its outcome. The outcome must be deterministic. Added test against current MainNet block height.

- burn projection model was totally wrong
- dex lp is `0.2545` not `0.2445`
- simplified `getBlockSubsidy` code


canonbrother (approved):

siradji (approved):
Much better!

PR #530, opened 2021-11-05
Add auction history get and list endpoints

### LINKED ISSUE #417: Loan Auctions API

Exposing loan auctions API, to be added into `LoanController.ts` at server with `Loan.ts` whale-api-client.

```
GET: /loan/auctions
GET: /loan/auctions/:id
```

/triage accepted
/priority important-soon
/area module-api

jingyi2811:
What kind of PR is this?:
/kind feature

Add auction history get and list endpoints

Fixes #417

Additional comments?:


jingyi2811:
If C++ side has fixed the pagination issue, rerun the failed CI test will pass.

src/module.api/loan.auctionhistory.controller.e2e.ts
```diff
@@ -0,0 +1,415 @@
...
+      })
+    }
+    )
+  })
+})
+
+it('should listAuctionHistory with pagination', async () => {
+  const first = await controller.listAuctionHistory({ size: 2 })
+  expect(first.data.length).toStrictEqual(2)
+  expect(first.page?.next).toStrictEqual(`${first.data[1].vaultId}${first.data[1].blockHeight}`)
```
ivan-zynesis, line 390: If I understand L375 correctly, each data is for 1 batch, there is possibility for missing batch?

eg actual history (let say descending ordered)
[
  <vaultA><1000>, // batch 0
  <vaultA><1000>, // batch 1
  <vaultA><1000> // batch 2
  <vaultA><999> // batch 0
]

page1 = `listAuctionHistory({ size: 2 })`
page2 = `listAuctionHistory({ size: 2, next: <vaultA><1000> })`
`<vaultA><1000><batch2>` will be excluded right?

@Jouzo @surangap

src/module.api/loan.controller.ts
```diff
@@ -185,6 +186,17 @@ export class LoanController {
     return await this.vaultService.listAuction(query)
   }
 
+  /**
+   * Paginate loan auction history.
+   *
+   * @param {PaginationQuery} query
+   * @return {Promise<ApiPagedResponse<LoanAuctionHistory>>}
+   */
+  @Get('/auctionhistory')
```
canonbrother, line 195:
```suggestion
  @Get('/auctions/history')
```

src/module.api/loan.vault.service.ts
```diff
@@ -196,6 +225,19 @@ export class LoanVaultService {
       interestRate: scheme.interestrate.toFixed()
     }
   }
+
+  private mapLoanAuctionHistory (detail: ListAuctionHistoryDetail): LoanAuctionHistory {
```
canonbrother, line 229: it doesn't need to map if `ListAuctionHistoryDetail` and `LoanAuctionHistory` is same..

canonbrother (commented):
file naming -> use `.` as delimiter
eg:
loan.auction.history.test.ts

fuxingloh:
> file naming -> use `.` as delimiter eg: loan.auction.history.test.ts

@canonbrother need your help with taking over this PR

fuxingloh:
I am not sure what's the issue but if we can correct the problem on whale side to get the correct behaviour we should just do it at this stage.

canonbrother:
> I am not sure what's the issue but if we can correct the problem on whale side to get the correct behaviour we should just do it at this stage.

ahh.. ya.. you're right.. its whale issue

src/module.api/loan.vault.service.ts
```diff
@@ -100,6 +101,37 @@ export class LoanVaultService {
...
+          const full = await this.client.loan.listAuctionHistory('all', { limit: limit + 30, maxBlockHeight })
+          const index = full.findIndex(f => f.vaultId === vaultId)
+          if (index > 0) {
+            const sliced = full.slice(index + 1, index + 1 + query.size)
+            return ApiPagedResponse.of(sliced, query.size, item => `${item.vaultId}${item.blockHeight}`)
+          }
+          return await loop(limit + 30) // loop with larger size call until get the vaultId
+        }
+        return await loop()
+      }
```
canonbrother, line 125: why?? because the filtered by `vaultId` and `index` don't work

fuxingloh: Oh damn, we can't do `/vaults/:vaultId/auctions/:batch/history`?

fuxingloh: On Front End we need `Vault -> Auction -> Batch -> History`, means we need to filter by `vaultId` then `batchId` and get many history.

canonbrother: yes the only method now is doing filter on our end
grab full list -> find the target vaultId batch -> return

fuxingloh: Yeah, damn you are right I just checked the LSM Key on ain.

fuxingloh: Could you index `AuctionBatchHistory` on the whale side, perhaps prioritize this task. The front end needs a performant endpoint, I don't think we can use the current implementation.

Issue #541, opened 2021-11-10
`DeFiCh/ain:2.0.x` Loan Indexing and API


canonbrother:
#### Loan Scheme
- [ ] setLoanScheme @canonbrother  #560 
- [ ] setDefaultLoanScheme @canonbrother #622 
- [ ] destroyLoanScheme @canonbrother #612

#### Collateral/Loan Token
- [ ] setCollateralToken @ivan-zynesis 
- [ ] setLoanToken @monstrobishi #588 
- [ ] setLoanToken (keep loan token history) @ivan-zynesis 
- [ ] updateLoanToken @ivan-zynesis 

#### Vault
- [ ] createVault @monstrobishi 
- [ ] updateVault @monstrobishi 
- [ ] depositToVault @canonbrother 
- [ ] takeLoan
- [ ] loanPayback @monstrobishi 
- [ ] withdrawFromVault 
- [ ] closeVault
- [ ] placeAuctionBid @canonbrother 

### API
- [ ] listLoanSchemes @canonbrother #627 
- [ ] getLoanScheme @canonbrother #627 
- [ ] listCollateralTokens
- [ ] getCollateralToken
- [ ] getInterest
- [ ] getLoanToken
- [ ] listLoanTokens
- [ ] getLoanInfo
- [ ] listAuctions @jingyi2811 
- [ ] listAuctionHistory @jingyi2811

fuxingloh:
Not all API/Indexing are important, to focus on `VaultHistory` first.

monstrobishi:
I'll take over some of the vault tasks

PR #560, opened 2021-11-12
Index `setLoanScheme`

canonbrother:

/kind feature

Index `setLoanScheme`

Fixes part of #541


src/module.model/loan.scheme.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  async query (limit: number, lt?: string): Promise<LoanScheme[]> {
+    return await this.database.query(LoanSchemeMapping.index.loan_scheme_id, {
+      limit: limit,
+      order: SortOrder.DESC,
+      lt: lt
+    })
+  }
+
```
ivan-zynesis, line 31: `getActiveLoanScheme (schemeId)` probably the most needed func, to get activated, search for a row activation height less than current height

src/module.indexer/rpc.block.provider.ts
```diff
@@ -136,6 +138,7 @@ export class RPCBlockProvider {
     try {
       await this.indexer.index(block)
       await this.statusMapper.put(hash, height, Status.INDEXED)
+      await this.daemon.process(block)
```
canonbrother, line 141: @fuxingloh i need a daemon process so i put it here first.. will do better structure in near while

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      await this.loanSchemeHistoryMapper.put({
+        id: `${data.identifier}-${block.height}`,
+        loanSchemeId: data.identifier,
+        sort: HexEncoder.encodeHeight(block.height),
+        ratio: data.ratio,
+        rate: new BigNumber(data.rate),
+        activateAfterBlock: data.update,
+        event: data.update.eq(0) ? LoanSchemeHistoryEvent.CREATE : LoanSchemeHistoryEvent.UPDATE,
```
ivan-zynesis, line 57: CREATE or UPDATE is depend if the scheme ID existed instead of the `activation block`

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,89 @@
...
+      await this.loanSchemePendingMapper.delete(data.identifier)
+      await this.loanSchemeHistoryMapper.delete(`${data.identifier}-${block.height}`)
+    }
+  }
+
+  /**
+   * Get previous loan scheme before current height
+   */
+  private async getPrevious (loanSchemeId: string): Promise<LoanSchemeHistory> {
+    const histories = await this.loanSchemeHistoryMapper.query(loanSchemeId, 1)
```
ivan-zynesis, line 82: Another challenging part, it should restore to last "activated" one.

src/module.indexer/model/daemon.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@nestjs/common'
+import { LoanSchemeMapper } from '@src/module.model/loan.scheme'
+import { LoanSchemePendingMapper } from '@src/module.model/loan.scheme.pending'
+import BigNumber from 'bignumber.js'
+import { blockchain as defid } from '@defichain/jellyfish-api-core'
+
+@Injectable()
+export class Daemon {
```
ivan-zynesis, line 8: What I mean by "daemon" simply saying it run on every block regardless of dftx type found in that raw block.
It still need `index` and `invalidate`, invalidate to undo the "delayed activation"

fuxingloh: What I hate about this is its performance implication since it's not linked to DfTx anymore. Hmmmm

ivan-zynesis: Agree, and this overhead is redundant more than 90% of the time. Another way is completely keep them in 1 table without separating history and normal, every row do come with height. `listLoanScheme` will have to be relayed to RPC like `listPoolPair`

canonbrother: great.. hmm.. yea.. this is my first naive solution.. my first thought was separate indexer and daemon functionality, i just parked the daemon inside the indexer runner temp..

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+    // get the closest activateAfterBlock against height
+    // ensure its queried by DESC height
+    // looking for height > activateHeight
+    const prevActiveLoanScheme = histories.find(h => new BigNumber(height).gte(h.activateAfterBlock))
+    if (prevActiveLoanScheme === undefined) {
+      throw new NotFoundIndexerError('index', 'LoanSchemeHistory', loanSchemeId)
+    }
+    return prevActiveLoanScheme
+  }
```
canonbrother, line 101: @ivan-zynesis 
the easiest way
eg: 
```
const height = 128
const histories = [
  { activateAfterBlock: 135 },
  { activateAfterBlock: 130 },
  { activateAfterBlock: 120 },
  { activateAfterBlock: 110 }
]
const latestActive = histories.find(h => height > h.activateAfterBlock) // 120
```

src/module.indexer/model/dftx/set.loan.scheme.ts
canonbrother: outdated
move to https://github.com/DeFiCh/whale/pull/560#discussion_r751197817

src/module.indexer/model/dftx/set.loan.scheme.ts
canonbrother: i did it based on the ain LoanScheme design
if loanScheme.update == 0 -> CREATE else UPDATE

btw your suggestion is right too

canonbrother: updated

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
```diff
@@ -0,0 +1,71 @@
...
+  constructor (
+    private readonly loanSchemeMapper: LoanSchemeMapper,
+    private readonly loanSchemeHistoryMapper: LoanSchemeHistoryMapper,
+    private readonly deferredLoanSchemeMapper: DeferredLoanSchemeMapper
+  ) {
+    super()
+  }
+
+  async index (block: RawBlock): Promise<void> {
+    const list = await this.deferredLoanSchemeMapper.query(100)
```
fuxingloh, line 25:
```suggestion
    const list = await this.deferredLoanSchemeMapper.query(Number.MAX_SAFE_INTEGER)
```

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
```diff
@@ -0,0 +1,71 @@
...
+    for (const each of list) {
+      if (new BigNumber(block.height).gte(each.activateAfterBlock)) {
+        await this.loanSchemeMapper.put(each)
+        await this.deferredLoanSchemeMapper.delete(each.id)
+      }
+    }
```
fuxingloh, line 31: I would sort deferred actions by height in ascending when first item in the list doesn't meet the requirement, I would reject all in the list.

fuxingloh: If you do this, it's better to paginate than get all at once since you will reject if conditions failed. But for this scheme setup, the implications are quite low due to low data traffic

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,108 @@
...
+    return loanScheme.update.eq(0) ||
+           loanScheme.update.eq(new BigNumber('0xffffffffffffffff')) ||
+           new BigNumber(height).gte(loanScheme.update)
+  }
+
+  /**
+   * Get previous active loan scheme
+   */
+  private async getPrevious (id: string, height: number): Promise<LoanSchemeHistory> {
+    const histories = await this.loanSchemeHistoryMapper.query(id, 50)
```
ivan-zynesis, line 95: 1. need while loop, could be last 50 all do not belongs to this scheme.
2. history length = 0 equal to delete scheme

ivan-zynesis (changes requested):

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
canonbrother: > I would sort deferred actions by height in ascending when first item in the list doesn't meet the requirement, I would reject all in the list.
the height doesn't matter as the deferLoanSchemeMapper just like `[s150, s200, s250, s300]`, i just pick and overwrite into loanSchemeMapper

> If you do this, it's better to paginate than get all at once since you will reject if conditions failed. But for this scheme setup, the implications are quite low due to low data traffic
yea.. pagination is required

canonbrother: > I would sort deferred actions by height in ascending when first item in the list doesn't meet the requirement, I would reject all in the list.

hmm.. rethink about it

canonbrother: gonna remap `deferredLoanSchemeMapper` to support sort key on height

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
```diff
@@ -0,0 +1,84 @@
...
+        await this.loanSchemeMapper.put(each)
+        await this.deferredLoanSchemeMapper.delete(each.id)
+      }
+      return await loop(activeAfterBlock, list[list.length - 1].block.height)
+    }
+
+    return await loop(block.height)
+  }
+
+  async invalidate (block: RawBlock, txns: Array<DfTxTransaction<LoanScheme>>): Promise<void> {
```
jingyi2811, line 41: I am thinking is there a way to test this `invalidate` function. I believe you should delete from loanSchemeMapper as you add an entry in https://github.com/DeFiCh/whale/pull/560/files#diff-7ff60895c2f24c1c872034f35e84df4124d83935c496fe3758b28a0cadb3669aR32

src/module.indexer/model/dftx/set.loan.scheme.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+        }
+      })
+
+      const list = await loanSchemeMapper.query(30)
+      expect(list).toStrictEqual([
+        {
+          id: 's250',
+          ratio: 255,
+          rate: '2.55',
+          activateAfterBlock: '18446744073709551615',
```
jingyi2811, line 193: If you want to use `'18446744073709551615'`, you probably should use `new BigNumber('0xffffffffffffffff')`. It looks much nicer

src/module.indexer/model/dftx/set.loan.scheme.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+  return loanSchemeId
+}
+
+async function updateLoanScheme (nameAsId: string, minColRatio: number, interestRate: BigNumber, activateAfterBlock?: number): Promise<string> {
+  const payload: any = {
+    id: nameAsId,
+    minColRatio: minColRatio,
+    interestRate: interestRate
+  }
+  if (activateAfterBlock !== undefined) {
```
jingyi2811, line 40: You might not want to add activateAfterBlock if activateAfterBlock == new BigNumber('0xffffffffffffffff')

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,116 @@
...
+        await this.loanSchemeMapper.put(previous)
+      } else {
+        await this.deferredLoanSchemeMapper.delete(data.identifier)
+      }
+
+      await this.loanSchemeHistoryMapper.delete(`${data.identifier}-${block.height}`)
+    }
+  }
+
+  private async has (id: string): Promise<boolean> {
```
jingyi2811, line 84: Not too sure you should add a function for this. If yes `has` maybe need to be replaced by a better name.

src/module.indexer/model/dftx/set.loan.scheme.ts
```diff
@@ -0,0 +1,116 @@
...
+    return !((await this.loanSchemeMapper.get(id)) == null)
```
jingyi2811, line 85:
```suggestion
    return (await this.loanSchemeMapper.get(id)) != null)
```

jingyi2811:
Quite good. Left some comments

src/module.indexer/model/dftx/set.loan.scheme.spec.ts
canonbrother: i don't what's wrong here.. its just a customize function on spec.ts

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
canonbrother: i have one previously but its removed.. anyway.. we don't a standardize way to test `invalidate` yet

src/module.indexer/model/dftx/set.loan.scheme.ts
canonbrother: please also give a suggestion as well if you have comment on it

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  async indexBlockStart (block: RawBlock): Promise<void> {
+    const loop = async (activeAfterBlock: number, next?: number): Promise<void> => {
+      const list = await this.deferredLoanSchemeMapper.query(activeAfterBlock, 100)
+      if (list.length === 0) {
+        return
+      }
+      for (const each of list) {
+        await this.loanSchemeMapper.put(this.mapLoanScheme(each))
+        await this.deferredLoanSchemeMapper.delete(each.id)
```
ivan-zynesis, line 33: Cannot really hard delete them.

at `t10`: set with activate height = `t20`
fork happen between t10 and t20

indexing up to 21
invalidate back to 19
index block 20 again, the deferred action will be gone.

src/module.indexer/model/dftx/set.deferred.loan.scheme.ts
```diff
@@ -0,0 +1,125 @@
...
+      rate: prevLoanScheme.rate,
+      activateAfterBlock: prevLoanScheme.activateAfterBlock,
+      block: prevLoanScheme.block
+    })
+  }
+
+  /**
+   * Get previous active loan scheme
+   */
+  private async getPrevLoanScheme (id: string, height: number): Promise<LoanSchemeHistory | undefined> {
```
ivan-zynesis, line 84: My strat in setCollateralToken indexer is add another index for `history` data and partitioned with `activated` or not, then can quickly retrieve last activated action.

canonbrother: @ivan-zynesis i know your strat but need it get approved first.. lemme follow the common implementation first.. its still not that bad

PR #573, opened 2021-11-15
Indexer: `setCollateralToken`

ivan-zynesis:

/kind feature

Fixes #


src/module.model/collateral.token.ts
```diff
@@ -0,0 +1,52 @@
...
+        key: (t: CollateralToken) => t.id
+      }
+    }
+  }
+}
+
+@Injectable()
+export class CollateralTokenMapper extends DeferableDftxMapper<CollateralToken> {
+  modelMapping = CollateralTokenMapping
+  historyMapping = undefined
```
ivan-zynesis, line 22: 3 mapper will be created automatically

base: using `modelMapping` defined here
history: using `historyMapping` (for extra query capability) + default mandatory indices created in the parent class
deferred action: automatically created by parent class only (operation purpose, no expected controller/api consuming it)

src/module.model/_deferrable.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * To find newest activated, below specific height, with highest block height
+   * example:
+   * A(t0): update with activation height = t4
+   * B(t1): update with activation height = t5
+   * C(t2): update with activation height = t4
+   * D(t3): update with activation height = t8
+   *
+   * return B if query execute at height t6
```
canonbrother, line 58: hmm.. shouldn't it 
```return B if query execute at height t5``` ??

File: src/module.indexer/model/dftx/_deferred.ts [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    super()
+  }
+
+  async indexTransaction (block: RawBlock, txns: DfTxTransaction<T>): Promise<void> {
+    const deferableAction = await this.mapDfTxToModel(txns.dftx.data, block)
+    const activateNow = deferableAction.activationHeight <= block.height
+
+    if (activateNow) {
+      await this.deferableDfTxMapper.getMapper().put(deferableAction)
```
canonbrother, line 21: need handle `delete` as well -> destroyLoanScheme

canonbrother: and `setDefaultLoanScheme`.. it will be needed to setAllFalsyDefault then set true only on the target loanScheme

fuxingloh:
🤔 There is some layered complexity to this, will revisit later to give a more effective review. Currently preoccupied with loan-related engineering.

ivan-zynesis:
> 🤔 There is some layered complexity to this, will revisit later to give a more effective review. Currently preoccupied with loan-related engineering.

Yes, have to add wildcard test case for base lib once we roughly think it approach is okay.

PR #574, opened 2021-11-15
Change symbol trimming based on FC height

### LINKED ISSUE #567: Change poolpair token symbol trimming between 8/16 chars based on FC height
Remove token symbol trimming to 8 characters long in token/poolpair indexers.

#### Why is this needed:
To have consistent behavior is post FC defid.
https://github.com/DeFiCh/ain/compare/v1.8.4...v2.0.0#diff-5b5f5b126aad5172f5e15066f2926f4afb57e75eb624b509f56233c14cd4b5c9R1040-R1042

monstrobishi:

/kind feature

Change symbol trimming based on FC height

Fixes #567


ivan-zynesis (approved):
The const ConsensusParams isn't very beautiful but seems the similar approach is found in masternode indexer too, not sure what's the better way of doing it, will just approve for now

fuxingloh (approved):

PR #576, opened 2021-11-15
Refactor `active-price` indexer + new tests

monstrobishi:

/kind chore

Added isLive false test to `active-price` indexer

Fixes #


fuxingloh (commented):
Does this check for feed that was active previously but has gone inactive?

jingyi2811 (approved):

fuxingloh (approved):

PR #606, opened 2021-11-19
Indexer: `SetLoanToken` / `UpdateLoanToken`

ivan-zynesis:

/kind feature

Index current loan token and loan token create/update history

Fixes #


src/module.indexer/model/dftx/set.loan.token.ts
```diff
@@ -5,38 +5,79 @@ import { Injectable, Logger } from '@nestjs/common'
...
       sort: HexEncoder.encodeHeight(tokenId),
-      symbol: data.symbol.trim().substr(0, MAX_TOKEN_SYMBOL_LENGTH),
-      name: data?.name?.trim().substr(0, MAX_TOKEN_NAME_LENGTH) ?? data.symbol.trim().substr(0, MAX_TOKEN_NAME_LENGTH),
+      symbol: symbol,
+      name: name,
       isDAT: true,
       isLPS: false,
       limit: new BigNumber(0).toFixed(8),
-      mintable: false,
+      mintable: data.mintable,
```
ivan-zynesis, line 49: I think this should respect the dftx value.

src/module.model/loan.token.history.ts
```diff
@@ -0,0 +1,80 @@
+import { Model, ModelMapping } from '@src/module.database/model'
+import { Injectable } from '@nestjs/common'
+import { Database, SortOrder } from '@src/module.database/database'
+
+const LoanTokenHistoryMapping: ModelMapping<LoanTokenHistory> = {
+  type: 'loan_token_history',
+  index: {
+    height_token_id: {
```
ivan-zynesis, line 8: Having 2 indices, thinking defiscan will may list all loan updates history without filtering.

src/module.model/loan.token.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export interface LoanToken extends Model {
+  id: string // -----------| 32 bytes id in hex
+  interest: string
+
+  // foreign table partition, to retrieve price feed (CRUD) history -> price feed
+  tokenCurrency: string // ---| tokenCurrencyMapper partition key
+  // FK, to retrieve latest token info
+  tokenId: string // ----------| loanTokenMapper partition key, tokenMapper id
```
ivan-zynesis, line 60: Need some opinions on these two. They kinda work like FK, hesitate for adding a snapshot here which can lead to multiple source of truth.

tokenCurrency = price ticker partition, lead to latest price ticker setting (eg weightage)
tokenId = token partition, lead to latest token name, symbol etc

src/module.indexer/model/dftx/update.loan.token.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const [lastVersion] = await this.loanTokenHistoryMapper.query(data.tokenTx, 1, `${block.height}`)
+
+    // restore token
+    await this.tokenMapper.put({
+      ...existingToken,
+      symbol: lastVersion.symbol,
+      name: lastVersion.name,
+      mintable: lastVersion.mintable
+      // block: createOrUpdate.block // TBD: this info can be referenced as token creation date
```
ivan-zynesis, line 78: Another one need opinion, keeping either created at or updated at is more intuitive?

src/module.indexer/model/dftx/set.loan.token.ts
monstrobishi: No we can't do that.

AIN trims it at the **consensus** level. This is written in consensus. We cannot align with node values if we don't trim here.

src/module.indexer/model/dftx/set.loan.token.ts
```diff
@@ -5,38 +5,79 @@ import { Injectable, Logger } from '@nestjs/common'
...
       decimal: 8,
       tradeable: true,
-      block: { hash: block.hash, height: block.height, medianTime: block.mediantime, time: block.time }
```
monstrobishi, line 34: No reason to change this imo, keep it consistent with other indexer implementation (Following Fuxing's pattern)

src/module.indexer/model/dftx/update.loan.token.ts
monstrobishi: I think keep created and just use the token creation value (Controller can handle that)

src/module.model/loan.token.ts
monstrobishi: For the comments on the end of ----| I think just keep it to the data structor only. You can keep the rest of the info in the // comment above

so maybe just:
```
// ---| token-currency (string)
// ---| tokenId (string)
```

src/module.indexer/model/dftx/set.loan.token.ts
canonbrother: although currently user can still set mintable: true via rpc cmd, but it will be forced to false in mainnet
https://github.com/DeFiCh/ain/blob/2baae1d961a72640cc1950bbaecc9b9537dbd521/src/masternodes/mn_checks.cpp#L694-L698

To reduce confusion, we result it `false` in whale layer better than client side

src/module.indexer/model/dftx/set.loan.token.ts
canonbrother: yea no need to change tho.. the previous one was more readable

src/module.indexer/model/dftx/set.loan.token.ts
```diff
@@ -5,38 +5,79 @@ import { Injectable, Logger } from '@nestjs/common'
...
+    }
+
+    // create new token
     await this.tokenMapper.put({
       id: `${tokenId}`,
       sort: HexEncoder.encodeHeight(tokenId),
-      symbol: data.symbol.trim().substr(0, MAX_TOKEN_SYMBOL_LENGTH),
-      name: data?.name?.trim().substr(0, MAX_TOKEN_NAME_LENGTH) ?? data.symbol.trim().substr(0, MAX_TOKEN_NAME_LENGTH),
+      symbol: symbol,
+      name: name,
```
canonbrother, line 45: add `isLoanToken` info in tokenMapper
https://github.com/DeFiCh/ain/blob/c7c15ea51b49b9f17687da9018a55bad3802e004/src/masternodes/rpc_tokens.cpp#L336

src/module.indexer/model/dftx/update.loan.token.ts
canonbrother: yes keep the `createdAt`, we have history to trace `updatedAt`

canonbrother (commented):
i'm good overall

src/module.indexer/model/dftx/set.loan.token.ts
monstrobishi: Oh I misread that section. Thought we were discussing the symbol and name. Didn't see that you moved it up. Please ignore my comment.

src/module.indexer/model/dftx/set.loan.token.ts
ivan-zynesis: Shall we open separate issue for that? it involve multiple test cases outside this indexer.

src/module.indexer/model/dftx/update.loan.token.ts
ivan-zynesis: That's why raise the question here to gather opinion.

Either way has own purpose, depend on what's expected to be presented on UI.

PR #622, opened 2021-11-24
Index `setDefaultLoanScheme`

canonbrother:

/kind feature

Fixes part of #541


src/module.model/loan.scheme.ts
```diff
@@ -47,6 +47,7 @@ export interface LoanScheme extends Model {
   ratio: number
   rate: BigNumber
   activateAfterBlock: BigNumber
+  default: boolean
```
ivan-zynesis, line 50: I thinking if there is better way of store this data. eg `LoanInfoDefaultMapper` for few reasons
- this is not highly deterministic where more than one scheme set true
- get default is actually quite inefficient, must loop through
- `default` is not really part of `LoanScheme`

Issue #310, opened 2021-08-03
Endpoint - `/stats/supply`


fuxingloh:
#### What would you like to be added:

Provide additional statistical data in `/stats` endpoint.

```json
{
  "supply": {
    "circulating": number,
    "burned": number,
    "total": number,
    "max": 1200000000
  }
}
```

- `supply.max` = 1,200,000,000
- `supply.burned` = Fee/Emission/Address Burns (basically all the burn from getBurnInfo)
- `supply.total` = Total Minted at block height
- `supply.circulating` = `supply.total` - `supply.burned`

https://academy.binance.com/en/glossary/circulating-supply

/triage accepted
/area module-api
/priority important-soon

fuxingloh:
/assign @siradji 

Do ping @monstrobishi for more details

siradji:
@monstrobishi, need your guide on this too. thanks

monstrobishi:
Assuming this is UTXO DFI, looking at the existing stats api it pulls it from the Bitcore node, indicating some indexing is required. To get an accurate value you might need to sum up all of the UTXO in the indexer, but I think Fuxing is suggesting here to calculate the emission up to the max years, and work backwards based on the block height.

I.e. "Total is 1.2 billion (based on whitepaper), every block produces ~190 blocks * the block height (that's circulating)"

For total you will want to read the section on the white paper on supply: https://defichain.com/white-paper/#token-issuance-schedule-via-staking

fuxingloh:
@siradji what happen to the PR for this?

joeldavidw:
@fuxingloh @siradji I will work on this

fuxingloh:
@ivan-zynesis I need your help with this, wanted to tackle this myself but this is much more complicated than I thought due to the evolution of rewards.

Initially, I did a deterministic calculation but it proves to be complex due to various historical evolution of rewards minting. Post Eunos hard fork, the reward distribution stabilized into a deterministic subsidy calculation.

My suggestion is we take a snapshot of up till let's say Fort Canning or Eunos HardFork. (By looping through all addresses that exist within DFC till a height.) And we derive a deterministic calculation from there onwards?

fuxingloh:
With https://github.com/DeFiCh/jellyfish/pull/1037 merged, we can use `BlockSubsidy` utility to get the running supply of DFI.

PR #661, opened 2021-12-16
`Indexer`: supply stats

ivan-zynesis:

/kind feature

Calculate and index DFI emission and burned amount on block basis

Fixes #

#### Additional comments?:
Would like to split adding /stats whale API as separate PR to ease review. Thus incomplete for #310


src/e2e.module.ts
```diff
@@ -74,7 +74,11 @@ export async function waitForIndexedHeight (app: NestFastifyApplication, height:
   const blockMapper = app.get(BlockMapper)
   await waitForExpect(async () => {
     const block = await blockMapper.getHighest()
-    await expect(block?.height).toBeGreaterThan(height)
+    if (height === 0) {
+      await expect(block?.height).toBeGreaterThanOrEqual(height)
```
ivan-zynesis, line 78: Only change to allow indexing genesis block, without the logic fork may break existing test

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,209 @@
...
+import { MasternodeStatsMapper } from '@src/module.model/masternode.stats'
+import { SupplyStatMapper } from '@src/module.model/supply.stat'
+import { SupplyStatAggregation, SupplyStatAggregationMapper } from '@src/module.model/supply.stat.aggregation'
+import { HexEncoder } from '@src/module.model/_hex.encoder'
+import BigNumber from 'bignumber.js'
+
+type EmissionRateBound = 'GENESIS' | 'POST_GENESIS' | 'EUNOS'
+
+@Injectable()
+export class EmissionCalculator {
```
ivan-zynesis, line 15: Not sure we park this into another unique module directory

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,209 @@
...
+          }
+          sum += burnedAmt
+        }
+      }
+    }
+    return sum
+  }
+
+  // TODO: extract by interpreting block data (another huge indexer) without relying on rpc
+  private async _extractBurnHistory (block: RawBlock): Promise<BurnHistory[]> {
```
ivan-zynesis, line 178: Expecting burn history listing RPC to handle all the operation burn, foundation nuke etc. Otherwise will have to work on full burn history by dftx decoding

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      const params = EmissionConsensusParams[this.network]
+      const reductionCount = Math.floor((currentHeight - eunosHeight) / params.EMISSION_REDUCTION_PERIOD)
+
+      let subsidy = new BigNumber(params.EUNOS_BASE_BLOCK_SUBSIDY)
+      for (let i = 0; i < reductionCount; i++) {
+        subsidy = subsidy.minus(subsidy.times(params.EMISSION_REDUCE_RATIO))
+      }
+      return subsidy.dp(8).toNumber()
```
surangap, line 41: @ivan-zynesis 
it's the same implementation as cpp side. you can find the full subsidy calculation here -> https://github.com/DeFiCh/ain/blob/c846338c05214b0745e8b63ec46c6b402fd6b0a2/src/validation.cpp#L1120
But this part is important.
```CAmount reductionAmount;
for (size_t i = reductions; i > 0; --i) {
    reductionAmount = (nSubsidy * consensusParams.emissionReductionAmount) / 100000;
    if (!reductionAmount) {
        nSubsidy = 0;
        break;
    }

    nSubsidy -= reductionAmount;
}
```
note the  `reductionAmount = (nSubsidy * consensusParams.emissionReductionAmount) / 100000;` since defid all calculations are on `CAmount` aka `int64_t` , there's going to be a precision loss in `reductionAmount` . so this can accumulate within the loop. Whereas in ts, with Bignumber, we have more accuracy. Therefore the final numbers might be slightly different. Might need a test to compare the results using both sides and confirm.

src/module.api/stats.controller.ts
```diff
@@ -40,7 +42,8 @@ export class StatsController {
       net: await this.cachedGet('net', this.getNet.bind(this), 1800),
       blockchain: {
         difficulty: block.difficulty
-      }
+      },
+      supply: await this.cachedGet('supply', this.getSupplyDetail.bind(this), 30)
```
ivan-zynesis, line 46: Caching only for 1 block long to get close to real time update.

src/module.indexer/model/supply.stat.ts
ivan-zynesis: Changed implementation to be in satoshi using BigInt (no more decimal/float, anything below unity will be dropped)

ivan-zynesis: > Therefore the final numbers might be slightly different

Was thinking about this too, if the calculation off by 1 satoshi every X blocks, after pile up for millions block the total stat will be off significantly. But will be very tricky to have test to index millions blocks.

canonbrother: > Was thinking about this too, if the calculation off by 1 satoshi every X blocks, after pile up for millions block the total stat will be off significantly. But will be very tricky to have test to index millions blocks. Any suggestion?

hmm... can store the base unit? easy to verify if any precision loss

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,210 @@
...
+  EUNOS_HEIGHT: number
+  EUNOS_BASE_BLOCK_SUBSIDY_SATOSHI: bigint
+  EMISSION_REDUCTION: bigint
+  EMISSION_REDUCTION_PERIOD: number
+}
+
+const EmissionConsensusParams: { [key in NetworkName]: EmissionConsensusParamsI } = {
+  mainnet: {
+    EUNOS_HEIGHT: 894000,
+    PRE_EUNOS_BASE_BLOCK_SUBSIDY: 200,
```
canonbrother, line 78: i think should standardize the base block subsidy unit as satoshi since the new one using satoshi
```suggestion
    PRE_EUNOS_BASE_BLOCK_SUBSIDY: BigInt(20000000000),
```

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,210 @@
...
+  testnet: {
+    EUNOS_HEIGHT: 354950,
+    PRE_EUNOS_BASE_BLOCK_SUBSIDY: 200,
+    EUNOS_BASE_BLOCK_SUBSIDY_SATOSHI: BigInt(40504000000),
+    EMISSION_REDUCTION: BigInt(1658), // 1.658%, 10000x magnified, as int
+    EMISSION_REDUCTION_PERIOD: 32690 // 2 weeks
+  },
+  regtest: {
+    EUNOS_HEIGHT: 50,
+    PRE_EUNOS_BASE_BLOCK_SUBSIDY: 200,
```
canonbrother, line 92:
```suggestion
    PRE_EUNOS_BASE_BLOCK_SUBSIDY: BigInt(10000000000),
```

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,210 @@
...
+    EUNOS_BASE_BLOCK_SUBSIDY_SATOSHI: BigInt(40504000000),
+    EMISSION_REDUCTION: BigInt(1658), // 1.658%, 10000x magnified, as int
+    EMISSION_REDUCTION_PERIOD: 5
```
canonbrother, line 95:
```suggestion
    EMISSION_REDUCTION_PERIOD: 32690
```

src/module.indexer/model/supply.stat.ts
canonbrother: hmm.. maybe need check the `-jellyfish_regtest` as well
they have diff value
```
consensus.baseBlockSubsidy = (isJellyfish) ? 100 * COIN : 50 * COIN;
```

src/module.indexer/model/supply.stat.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  async index (block: RawBlock): Promise<void> {
+    const lastAggregated: SupplyStatData = block.height === 0
+      ? {
+          circulating: 0,
+          burned: 0,
+          locked: 0,
+          total: 0
+        }
+      : (await this.aggregation.query(1, HexEncoder.encodeHeight(block.height)))[0]
```
canonbrother, line 124: preferably create another getLatest() for this

src/module.indexer/model/supply.stat.ts
canonbrother: a little refactor.. maybe move it to API section?

canonbrother: IMO.. its too complex to index burn history...

canonbrother (commented):
overall good for me.. the requests doesn't affect the core result

src/module.indexer/model/supply.stat.ts
ivan-zynesis: Storing in number should be fine. 1.00000001 DFI and 100000001 satoshi not likely make any difference. Keep data in bigint have many compatibility issue, eg module.database. The precision issue came from the calculation process instead of storage.

src/module.indexer/model/supply.stat.ts
ivan-zynesis: Will completely drop this, not used, no calculation needed for pre Eunos emission anyway.

src/module.indexer/model/supply.stat.ts
ivan-zynesis: Small reduction period was set intentionally for test purpose. Other wise we need it to index to 70k+ block to assert the new subsidy amount.

ivan-zynesis: `regtest` is standalone in each machine, whatever is used in defid regtest, doesn't affect the test cases here.

src/module.indexer/model/supply.stat.ts
ivan-zynesis: Is okay, test cases here is independent on how much defid actually emitted. Numbers here are solely for calculation only regardless the env. Very difficult to use actual regtest consensus params for testing (calculate the expected total emission), the reduction only happen every 2 weeks of block.

PR #666, opened 2021-12-17
add `ListAccountHistory` API

### LINKED ISSUE #660: exposing `/address/:address/history` endpoint

Map `/address/:address/history` -> `listAccountHistory` with data model and interface remapped.

/triage accepted
/assign @canonbrother 
/area module-api

Related to:
- https://github.com/DeFiCh/wallet/issues/1732
- https://github.com/DeFiCh/scan/issues/669
- https://github.com/DeFiCh/whale/pull/367

canonbrother:

/kind feature

Fixes #660

### DIFF
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -49,6 +49,7 @@ export class AddressController {
 
     if (next !== undefined) {
       const [txid, txType, maxBlockHeight] = next.split('-')
+
       const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
         const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
         if (list.length === 0) {
@@ -57,14 +58,14 @@ export class AddressController {
         const foundIndex = list.findIndex(each => each.txid === txid && each.type === txType)
         if (foundIndex === -1) {
           // if not found, extend the size till grab the 'next'
-          return await loop(list[list.length - 1].blockHeight, limit * 2)
+          return await loop(Number(maxBlockHeight), limit * 2)
         }
         const start = foundIndex + 1 // plus 1 to exclude the prev txid
         const size = start + query.size
         const sliced = list.slice(start, size)
         if (sliced.length !== query.size) {
           // need a bigger volume to achieve the size
-          return await loop(list[list.length - 1].blockHeight, limit * 2)
+          return await loop(Number(maxBlockHeight), limit * 2)
         }
         return sliced
       }
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -43,19 +43,29 @@ export class AddressController {
       throw new ForbiddenException('mine is not allowed')
     }
 
-    const limit = query.size > 30 ? 30 : query.size
+    const limit = query.size > 100 ? 100 : query.size
     const next = query.next ?? undefined
     let list: AccountHistory[]
 
     if (next !== undefined) {
       const [txid, txType, maxBlockHeight] = next.split('-')
-      // NOTE(canonbrother): filter refers to block, size refers to tx, block has many txs
-      // set limit at depth for easier
-      list = await this.rpcClient.account.listAccountHistory(address, { limit: Number.MAX_SAFE_INTEGER, maxBlockHeight: Number(maxBlockHeight), depth: limit })
-      const prev = list.findIndex(each => each.txid === txid && each.type === txType)
-      const start = prev + 1 // plus 1 to exclude the prev txid
-      const size = start + limit
-      list = list.slice(start, size)
+      const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
+        const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
+        const foundIndex = list.findIndex(each => each.txid === txid && each.type === txType)
+        if (foundIndex === -1) {
+          // if not found, extend the size till grab the 'next'
+          return await loop(list[list.length - 1].blockHeight, limit * 2)
+        }
+        const start = foundIndex + 1 // plus 1 to exclude the prev txid
+        const size = start + query.size
+        const sliced = list.slice(start, size)
+        if (sliced.length !== query.size) {
+          // need a bigger volume to achieve the size
+          return await loop(list[list.length - 1].blockHeight, limit * 2)
+        }
+        return sliced
+      }
+      list = await loop(Number(maxBlockHeight), limit)
     } else {
       list = await this.rpcClient.account.listAccountHistory(address, { limit: limit })
     }
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -43,7 +43,7 @@ export class AddressController {
       throw new ForbiddenException('mine is not allowed')
     }
 
-    const limit = query.size > 100 ? 100 : query.size
+    const limit = query.size > 200 ? 200 : query.size
     const next = query.next ?? undefined
     let list: AccountHistory[]
 
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -51,6 +51,9 @@ export class AddressController {
       const [txid, txType, maxBlockHeight] = next.split('-')
       const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
         const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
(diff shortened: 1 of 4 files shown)


src/module.api/address.controller.ts
```diff
@@ -30,6 +31,51 @@ export class AddressController {
...
+   * @param {PaginationQuery} query
+   */
+  @Get('/history')
+  async listAccountHistory (
+    @Param('address') address: string,
+      @Query() query: PaginationQuery): Promise<ApiPagedResponse<AddressHistory>> {
+    const limit = query.size > 30 ? 30 : query.size
+    const next = query.next ?? undefined
+    let list: AccountHistory[]
+
```
ivan-zynesis, line 45: Can straight reject if `address` = `mine`

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,51 @@ export class AddressController {
...
+    if (next !== undefined) {
+      const [txid, txType, maxBlockHeight] = next.split('-')
+      // NOTE(canonbrother): filter refers to block, size refers to tx, block has many txs
+      // set limit at depth for easier
+      list = await this.rpcClient.account.listAccountHistory(address, { maxBlockHeight: Number(maxBlockHeight), depth: limit })
```
ivan-zynesis, line 50: `depth` is counting by block

ivan-zynesis, line 50: I guess using `limit` will be easier than `depth`
depth is by block, the length is unpredictable.

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,51 @@ export class AddressController {
...
+  @Get('/history')
+  async listAccountHistory (
+    @Param('address') address: string,
+      @Query() query: PaginationQuery): Promise<ApiPagedResponse<AddressHistory>> {
+    const limit = query.size > 30 ? 30 : query.size
+    const next = query.next ?? undefined
+    let list: AccountHistory[]
+
+    if (next !== undefined) {
+      const [txid, txType, maxBlockHeight] = next.split('-')
```
ivan-zynesis, line 47: `txid-n-height` will be unique.

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,51 @@ export class AddressController {
...
 
+  /**
+   * @param {string} address to list participate account history
+   * @param {PaginationQuery} query
+   */
+  @Get('/history')
+  async listAccountHistory (
+    @Param('address') address: string,
+      @Query() query: PaginationQuery): Promise<ApiPagedResponse<AddressHistory>> {
+    const limit = query.size > 30 ? 30 : query.size
```
ivan-zynesis, line 42: Have reject large `query.size` to prevent attack.

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,51 @@ export class AddressController {
...
+      list = await this.rpcClient.account.listAccountHistory(address, { maxBlockHeight: Number(maxBlockHeight), depth: limit })
+      const prev = list.findIndex(each => each.txid === txid && each.type === txType)
+      const start = prev + 1 // plus 1 to exclude the prev txid
+      const size = start + limit
+      list = list.slice(start, size)
+    } else {
+      list = await this.rpcClient.account.listAccountHistory(address, { limit: limit })
+    }
+
+    const history: AddressHistory[] = list.map(each => {
```
fuxingloh, line 59: separate into another function

src/module.api/address.controller.ts
fuxingloh, line 42: let's remove max 30 and use the default query.size max since it's indexed

fuxingloh (commented):
the pagination seems really tricky as you mentioned 🤔🤔🤔


> txn is used as part of the key and we don't have a way to pass this in.
```cpp
CScript owner;
uint32_t blockHeight;
uint32_t txn;
```

src/module.api/address.controller.ts
canonbrother: i was using limit at `limit`.. that's required many conditions write up and will be required number of rpc calls
so here's the comparison

[limit on depth]
1. call once -> filter -> done (fortunately it can be limited by owner address and depth)

[limit on limit]
2. requires number of call if can't grab the "next"
number of call to increase the size to get the "next", if next found, slice it, if the sliced's length == query.size then return else make another call
maybe i'm still not come out a good solution on this :thinking:

src/module.api/address.controller.ts
canonbrother: hmm... do you mean throw error for large size instead of silently default??

ivan-zynesis: My bad, I thought it was `query.size === undefined ? : 30 : query.size`
I am okay with capped `size`

src/module.api/address.controller.ts
ivan-zynesis: hmm valid concern. Indeed, `maxBlockHeight` + `depth` is easier implementation but each "paged" response length will be very random

src/module.api/address.controller.ts
canonbrother: i was thinking to remove the limit on jellyfish rpc
https://github.com/DeFiCh/jellyfish/blob/ed1f9782889a2bb52e938e3ddb6ac6799d11d654/packages/jellyfish-api-core/src/category/account.ts#L297

or create another `listAccountHistoryWithoutLimit` ??

src/module.api/address.controller.ts
canonbrother: its good to use `n`.. however i just use what's resp provided, it doesn't include tx.n... might need another map index process

src/module.api/address.controller.ts
canonbrother: understand.. i'm limit 30 depths while has 'next' provided
30 blocks is sufficient

src/module.api/address.controller.ts
canonbrother: the length should be always fixed, the filter will grab the size exactly

canonbrother:
@ivan-zynesis  @fuxingloh
for easier
i refine a "limit call" version for you reference

either go for the prev approach "volume call" (volume data) or "limit call" (number of rpc calls)

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,52 @@ export class AddressController {
...
+
+    const limit = query.size > 200 ? 200 : query.size
+    const next = query.next ?? undefined
+    let list: AccountHistory[]
+
+    if (next !== undefined) {
+      const [txid, txType, maxBlockHeight] = next.split('-')
+      const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
+        const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
+        const foundIndex = list.findIndex(each => each.txid === txid && each.type === txType)
```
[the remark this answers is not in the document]
canonbrother: unfortunately.. not doable.. the resp doesn't provide `n`.. so i do a remap with n but will have issue while the next.blockHeight is changed.. whole manual mapped index will gonna change

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,55 @@ export class AddressController {
...
+      const [txid, txType, maxBlockHeight] = next.split('-')
+      const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
+        const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
+        if (list.length === 0) {
+          return []
+        }
+        const foundIndex = list.findIndex(each => each.txid === txid && each.type === txType)
+        if (foundIndex === -1) {
+          // if not found, extend the size till grab the 'next'
+          return await loop(list[list.length - 1].blockHeight, limit * 2)
```
fuxingloh, line 60:
```suggestion
          return await loop(Number(maxBlockHeight), limit * 2)
```

src/module.api/address.controller.ts
```diff
@@ -30,6 +31,55 @@ export class AddressController {
...
+        }
+        const start = foundIndex + 1 // plus 1 to exclude the prev txid
+        const size = start + query.size
+        const sliced = list.slice(start, size)
+        if (sliced.length !== query.size) {
+          // need a bigger volume to achieve the size
+          return await loop(list[list.length - 1].blockHeight, limit * 2)
```
fuxingloh, line 67:
```suggestion
          return await loop(Number(maxBlockHeight), limit * 2)
```

fuxingloh (commented):
should these 2 use the current provided maxBlockHeight instead of getting the last one?

Hmm, maybe for this recursive condition the last `blockHeight` is also the current `maxBlockHeight`?

fuxingloh (approved):

PR #675, opened 2021-12-23
Remove rewards from `listaccounthistory` API

ivan-zynesis:

/kind fix

Pagination implementation do not work, will crash when call listing with `next` due to undefined txid for reward account history

Fixes #


src/module.api/address.controller.ts
```diff
@@ -51,7 +51,11 @@ export class AddressController {
       const [txid, txType, maxBlockHeight] = next.split('-')
 
       const loop = async (maxBlockHeight: number, limit: number): Promise<AccountHistory[]> => {
-        const list = await this.rpcClient.account.listAccountHistory(address, { limit: limit, maxBlockHeight: maxBlockHeight })
+        const list = await this.rpcClient.account.listAccountHistory(address, {
+          limit: limit,
+          maxBlockHeight: maxBlockHeight,
+          no_rewards: true
```
fuxingloh, line 57: I believe we need to add for line 78 too?

fuxingloh (approved):

PR #699, opened 2022-01-13
dex hide BURN-DFI

fuxingloh:

/kind chore

Hide internal PoolPair BURN-DFI

### DIFF
--- a/src/module.api/poolpair.controller.ts
+++ b/src/module.api/poolpair.controller.ts
@@ -36,6 +36,10 @@ export class PoolPairController {
 
     const items: PoolPairData[] = []
     for (const [id, info] of Object.entries(result)) {
+      if (info.symbol === 'BURN-DFI') {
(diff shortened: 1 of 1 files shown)


src/module.api/poolpair.controller.ts
```diff
@@ -36,6 +36,10 @@ export class PoolPairController {
 
     const items: PoolPairData[] = []
     for (const [id, info] of Object.entries(result)) {
+      if (info.symbol === 'BURN-DFI') {
```
canonbrother, line 39: maybe hide those `status === false`

canonbrother (approved):

src/module.api/poolpair.controller.ts
fuxingloh: Probably not, for status === false, it's still shown on Light Wallet with trading disabled.

PR #713, opened 2022-01-18
feat(indexer): `poolswap` "lite"

### LINKED ISSUE #705: Basic DEX PoolSwap & CompositeSwap Indexing

Simplified indexing of PoolSwap and CompositeSwap to track volume and activity. For Volume and APR tracking only hence Add/Remove Pool DfTx is not required.

/triage accepted
/area module-indexer module-database
/priority important-soon

monstrobishi:

/kind feature

Light implementation calculates volume and commission. Commission apr is added to total apr.

Fixes #705


src/module.indexer/model/dftx/create.poolpair.ts
```diff
@@ -78,17 +79,17 @@ export class CreatePoolPairIndexer extends DfTxIndexer<PoolCreatePair> {
...
+      poolPairId: tokenId,
       block: { hash: block.hash, height: block.height, medianTime: block.mediantime, time: block.time }
     })
 
     await this.tokenMapper.put({
       id: `${tokenId}`,
       sort: HexEncoder.encodeHeight(tokenId),
       symbol: data.pairSymbol,
-      name: `${data.tokenA}-${data.tokenB} LP Token`,
+      name: `${tokenA.symbol}-${tokenA.symbol} LP Token`,
```
canonbrother, line 92:
```suggestion
      name: `${tokenA.symbol}-${tokenB.symbol} LP Token`,
```

src/module.indexer/model/dftx/compositeswap.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      await this.poolSwapIndexer.indexSwap(poolPair, poolSwap.fromTokenId, poolSwap.fromAmount, block, transaction.txn.txid)
+    }
+  }
+
+  async invalidateTransaction (block: RawBlock, transaction: DfTxTransaction<CompositeSwap>): Promise<void> {
+    const data = transaction.dftx.data
+    const poolSwap = data.poolSwap
+    const poolIds = data.pools
+    if (poolIds.length === 0) {
```
kodemon, line 53: `question:` wanted to gather some insights on value extraction in this method and other methods similar to it.

What is our thoughts around object spread, eg.

```ts
const { poolSwap, pools } = transaction.dftx.data
```

A quick secondary `nit:` is use of `poolIds` > `pools`. If `pools` is an array of objects, would `for (const pool of pools)` later in this function be more suitable?

I may be missing some underlying knowledge here so this is a shallow inquiry to expand my understanding. 🙏🏻

canonbrother (dismissed):

kodemon (approved):

src/module.api/poolpair.service.ts
```diff
@@ -117,6 +124,64 @@ export class PoolPairService {
...
+      if (dfiPair !== undefined) {
+        const dfiPrice = await this.getUSD_PER_DFI() ?? 0
+        if (dfiPair.idTokenA === '0') {
+          return dfiPair.reserveA.div(dfiPair.reserveB).times(dfiPrice)
+        } else if (dfiPair.idTokenB === '0') {
+          return dfiPair.reserveB.div(dfiPair.reserveA).times(dfiPrice)
+        }
+      }
+
+      const dusdPair = await this.getPoolPair(token, 'DUSD')
```
ivan-zynesis, line 150: can reverse the order, prioritize XXX-DUSD, avoid using any XXX-DFI, DFI-DUSD (I think do not exist at the moment)

src/module.api/poolpair.service.ts
```diff
@@ -117,6 +124,64 @@ export class PoolPairService {
...
+    const block = await this.blockMapper.getHighest()
+    const height = block?.height ?? 0
+    return await this.cache.get<PoolPairData['volume']>(`H24_VOLUME_${id}`, async () => {
+      const swaps = await this.poolSwapMapper.query(`${id}`, Number.MAX_SAFE_INTEGER, undefined,
+        HexEncoder.encodeHeight(height - 2880))
+
+      let accum = new BigNumber(0)
+      for (const swap of swaps) {
+        const amount = new BigNumber(swap.fromAmount)
+        accum = accum.plus(amount.times(await this.getPriceForToken(swap.fromTokenId) ?? 0))
```
ivan-zynesis, line 174: Can move the price retrieval out of for loop, in case cache expiry during summation.

src/module.indexer/model/dftx/poolswap.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (poolPairToken === undefined) {
+      throw new IndexerError(`Pool for pair ${data.fromTokenId}, ${data.toTokenId} not found`)
+    }
+
+    const poolPair = await this.poolPairMapper.getLatest(`${poolPairToken.poolPairId}`)
+    if (poolPair === undefined) {
+      throw new IndexerError(`Pool with id ${poolPairToken.poolPairId} not found`)
+    }
+
+    await this.poolPairMapper.delete(`${poolPair.poolPairId}-${block.height}`)
```
ivan-zynesis, line 69:
```suggestion
    await this.poolSwapMapper.delete(`${poolPair.poolPairId}-${block.height}`)
```

src/module.indexer/model/dftx/poolswap.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    if (poolPairToken === undefined) {
+      throw new IndexerError(`Pool for pair ${data.fromTokenId}, ${data.toTokenId} not found`)
+    }
+
+    const poolPair = await this.poolPairMapper.getLatest(`${poolPairToken.poolPairId}`)
+    if (poolPair === undefined) {
+      throw new IndexerError(`Pool with id ${poolPairToken.poolPairId} not found`)
+    }
+
```
ivan-zynesis, line 68: Can add a height check for found poolswap row

ivan-zynesis (changes requested):

src/module.indexer/model/dftx/compositeswap.ts
monstrobishi: Actually that's fair, I'll keep in mind for my next PR, but I think this can get through

src/module.indexer/model/dftx/poolswap.ts
monstrobishi: This is looking for poolpair row not poolswap row

src/module.api/poolpair.service.ts
monstrobishi: Was about to make this change, but unfortunately there is no way to know which token IDs we need prices for without looping over all the swaps.

In that case it would just be the same loop first without the `accum.plus`, which I don't think is a big computational load on modern machines.

I think leaving it as is would be for best.

src/module.api/poolpair.service.ts
monstrobishi: Can you elaborate on this, are you saying put the `TSLA-DUSD` test first?

monstrobishi: So is the rationale that there will be more XXX-DUSD pairs so we should check those first?

I'm not sure what the second part of the statement means

monstrobishi: Ok I've made this change

src/module.api/poolpair.service.ts
kodemon: In the future if we want to avoid sequential locks on price for token requests we can do a `Promise.all` to accumulate all swaps in parallel. But not sure if that is feasible in this particular instance or not, but definitely a future _potential_ optimization.

monstrobishi: I find `Promise.all` not super useful in these accumulator style cases

src/module.indexer/model/dftx/poolswap.ts
monstrobishi: Ok I got what you mean, but since txid is part of the block I think it's fine for this case, we can revisit when we refactor into jellyfish

ivan-zynesis (approved):

kodemon (dismissed):

src/module.api/poolpair.service.ts
```diff
@@ -117,6 +124,63 @@ export class PoolPairService {
...
+      ttl: 3600 // 60 minutes
+    })
+  }
+
+  public async getUSDVolume (id: string): Promise<PoolPairData['volume'] | undefined> {
+    const block = await this.blockMapper.getHighest()
+    const height = block?.height ?? 0
+    return await this.cache.get<PoolPairData['volume']>(`H24_VOLUME_${id}`, async () => {
+      const swaps = await this.poolSwapMapper.query(`${id}`, Number.MAX_SAFE_INTEGER, undefined,
+        HexEncoder.encodeHeight(height - 2880))
```
canonbrother, line 168: ahh forgot this.. do we need diff main and reg blocksPerDay on query??

src/module.indexer/model/dftx/compositeswap.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    for (const pool of poolIds) {
+      const poolPair = await this.poolPairMapper.getLatest(`${pool.id}`)
+
+      if (poolPair === undefined) {
+        throw new IndexerError(`Pool with id ${pool.id} not found`)
+      }
+
+      await this.poolSwapIndexer.indexSwap(poolPair, poolSwap.fromTokenId, poolSwap.fromAmount, block, transaction.txn.txid)
```
fuxingloh, line 45: There are multiple commissions after traversing through the composite path, should we `fromAmount * (1-commision)` to get a more accurate amount?

src/module.api/poolpair.service.ts
```diff
@@ -219,11 +283,15 @@ export class PoolPairService {
 
     const yearlyUSD = customUSD.plus(pctUSD).plus(loanUSD)
     // 1 == 100%, 0.1 = 10%
-    const apr = yearlyUSD.div(totalLiquidityUSD).toNumber()
+    const apr = yearlyUSD.div(totalLiquidityUSD)
+
+    const volume = await this.getUSDVolume(id)
+    const commission = info.commission.times(volume?.h24 ?? 0).times(365).div(totalLiquidityUSD)
```
fuxingloh, line 289: This is quite a heavily called API in ocean, I would not implement this until we have the bucketed rolling aggregator for performance sake?

fuxingloh (commented):
Did an initial review

src/module.api/poolpair.service.ts
ivan-zynesis: Thought about this issue, but the implementation itself wouldn't be lite any more. Will a 8 to 24 hours cache durability work?

monstrobishi: The db scan here shouldn't pose much of an issue, it's quite similar to how we're doing the oracle graphs. Plus it's cached.

src/module.indexer/model/dftx/compositeswap.ts
```diff
@@ -35,14 +36,22 @@ export class CompositeSwapIndexer extends DfTxIndexer<CompositeSwap> {
       poolIds.push({ id: poolPairToken.poolPairId })
     }
 
+    let swapAmount
```
kodemon, line 39:
```suggestion
    let swapAmount: BigNumber
```
Since there is no type by reference here, should we explicitly add it here?

src/module.api/poolpair.service.ts
```diff
@@ -219,11 +283,15 @@ export class PoolPairService {
 
     const yearlyUSD = customUSD.plus(pctUSD).plus(loanUSD)
     // 1 == 100%, 0.1 = 10%
-    const apr = yearlyUSD.div(totalLiquidityUSD).toNumber()
+    const apr = yearlyUSD.div(totalLiquidityUSD)
```
ivan-zynesis, line 228: Consider rename this to be not misleading, nothing to do with apr here.

src/module.api/poolpair.service.ts
ivan-zynesis: Hmm, * 365 can arguably a valid projection for a year, but I doubt should we do this. Need more opinions. @DeFiCh/jellyfish

monstrobishi: This section of the code was removed

src/module.indexer/model/dftx/compositeswap.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      if (poolPairToken === undefined) {
+        throw new IndexerError(`Pool for pair ${poolSwap.fromTokenId}, ${poolSwap.toTokenId} not found`)
+      }
+
+      poolIds.push({ id: poolPairToken.poolPairId })
+    }
+
+    for (const pool of poolIds) {
+      await this.poolPairMapper.delete(`${pool.id}-${block.height}`)
```
monstrobishi, line 73: Wait how did I miss this, hold off until this is fixed

monstrobishi: Ok that's fixed

src/module.api/poolpair.service.ts
monstrobishi: This code section was removed

kodemon (approved):

canonbrother (dismissed):

ivan-zynesis (approved):

src/module.model/poolswap.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  async delete (id: string): Promise<void> {
+    return await this.database.delete(PoolSwapMapping, id)
+  }
+}
+
+export interface PoolSwap extends Model {
+  id: string // ---------| poolpairId-txid
+  key: string // --------| poolpairId
+  sort: string // -------| height-txid
```
fuxingloh, line 49: Would `height-txnNo` be better? Since `txid` is not ordered based on block order.

fuxingloh (approved):

PR #719, opened 2022-01-20
Anchors controller

siradji:
Signed-off-by: Suraj Auwal <siradjiawoual@gmail.com>

/kind

Fixes #

### DIFF
--- a/src/module.api/anchors.controller.e2e.ts
+++ b/src/module.api/anchors.controller.e2e.ts
@@ -1,0 +1,155 @@
+import { MasterNodeRegTestContainer } from '@defichain/testcontainers'
+import { RegTestFoundationKeys } from '@defichain/jellyfish-network'
+import { NestFastifyApplication } from '@nestjs/platform-fastify'
+import { createTestingApp } from '@src/e2e.module'
+import { AnchorsController } from '@src/module.api/anchors.controller'
+import { TestingGroup } from '@defichain/jellyfish-testing'
+
+let tGroup: TestingGroup
+let container: MasterNodeRegTestContainer
+let app: NestFastifyApplication
+let controller: AnchorsController
+
+beforeAll(async () => {
+  tGroup = TestingGroup.create(3)
+  container = tGroup.group.get(0)
+
+  await container.start()
+  await tGroup.start()
+
+  app = await createTestingApp(container)
+  controller = app.get(AnchorsController)
+
+  await setup()
+})
+
+afterAll(async () => {
+  await app.close()
+  await tGroup.group.stop()
+})
+
+async function setMockTime (offsetHour: number): Promise<void> {
+  await tGroup.exec(async (testing: any) => {
+    await testing.misc.offsetTimeHourly(offsetHour)
+  })
+}
+
+async function setup (): Promise<void> {
+  {
+    const auths = await tGroup.get(0).container.call('spv_listanchorauths')
+    expect(auths.length).toStrictEqual(0)
+  }
+
+  const initOffsetHour = -12
+  await setMockTime(initOffsetHour)
+
+  for (let i = 0; i < 15; i += 1) {
+    const { container } = tGroup.get(i % tGroup.length())
+    await container.generate(1)
+    await tGroup.waitForSync()
+  }
+
+  await tGroup.get(0).container.waitForAnchorTeams(tGroup.length())
+
+  for (let i = 0; i < tGroup.length(); i += 1) {
+    const { container } = tGroup.get(i % tGroup.length())
+    const team = await container.call('getanchorteams')
+    expect(team.auth.length).toStrictEqual(tGroup.length())
+    expect(team.confirm.length).toStrictEqual(tGroup.length())
+    expect(team.auth.includes(RegTestFoundationKeys[0].operator.address))
+    expect(team.auth.includes(RegTestFoundationKeys[1].operator.address))
+    expect(team.auth.includes(RegTestFoundationKeys[2].operator.address))
+    expect(team.confirm.includes(RegTestFoundationKeys[0].operator.address))
+    expect(team.confirm.includes(RegTestFoundationKeys[1].operator.address))
+    expect(team.confirm.includes(RegTestFoundationKeys[2].operator.address))
+  }
+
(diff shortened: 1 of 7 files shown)


src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   *  List anchors
+   */
+
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery): Promise<ApiPagedResponse<AnchorData>> {
+    const result = await this.rpcClient.spv.listAnchors({
+      limit: Number(query.size),
+      ...(query.next !== undefined) && { startBTCHeight: Number(query.next) }
```
canonbrother, line 24: the `spv_listachors` order is from **tip to genesis**
so we can't use `startBTCHeight` as next query

```suggestion
      ...(query.next !== undefined) && { startBTCHeight: Number(query.next) }
      ...(query.next !== undefined) && { maxBtcHeight: Number(query.next) - 1 }
```

siradji: Not sure if this is gonna work for Pagination. 
https://github.com/DeFiCh/whale/pull/389#discussion_r750995359

siradji: @canonbrother, your suggestion  `from tip to genesis` help out. Thanks!

siradji:
Adding on top of what @canonbrother mentioned [here](https://github.com/DeFiCh/whale/pull/719#discussion_r788712357) , using `startBtcHeight` for pagination won't work.  Got it to work using `limit` and `maxBtcHeight`

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+export class AnchorsController {
+  constructor (
+    protected readonly rpcClient: JsonRpcClient
+  ) {
+  }
+
+  /**
+   *  List anchors
+   */
+
```
canonbrother, line 18:
```suggestion
```

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+  @Get('')
+  async list (
+    @Query() query: PaginationQuery): Promise<ApiPagedResponse<AnchorData>> {
+    const result = await this.rpcClient.spv.listAnchors({
+      limit: Number(query.size),
```
canonbrother, line 23: we need to have max limit 30 by default

src/module.api/anchors.controller.e2e.ts
```diff
@@ -0,0 +1,163 @@
...
+      previousAnchor: '0000000000000000000000000000000000000000000000000000000000000000',
+      rewardAddress: expect.any(String),
+      signatures: 2,
+      active: false,
+      anchorCreationHeight: 90
+    })
+  })
+
+  it('should test pagination with maxBtcHeight', async () => {
+    const firstRequest = await controller.list({ size: 3 })
```
canonbrother, line 150: need to elaborate more on pagination.. can help do `size: 1`?
like this
first
second
third
forth/last

src/module.api/anchors.controller.ts
siradji: Already done that here 
https://github.com/DeFiCh/whale/blob/547df776110f13fd211b8cc764822cb8090b9202/packages/whale-api-client/src/api/anchors.ts#L18-L21

kodemon: is this not handled by https://github.com/DeFiCh/whale/blob/main/src/module.api/_core/api.query.ts#L17 ?

src/module.api/anchors.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+
+@Controller('/anchors')
+export class AnchorsController {
+  constructor (
+    protected readonly rpcClient: JsonRpcClient
+  ) {
+  }
+
+  /**
+   *  List anchors
```
ivan-zynesis, line 16: Can add description for `query.next` is actually btc block height

ivan-zynesis (approved):

src/module.api/anchors.controller.e2e.ts
```diff
@@ -0,0 +1,177 @@
...
+    const { container } = tGroup.get(i % tGroup.length())
+    const auths = await container.call('spv_listanchorauths')
+    expect(auths.length).toStrictEqual(2)
+    expect(auths[0].signers).toStrictEqual(tGroup.length())
+  }
+
+  await tGroup.get(0).container.call('spv_setlastheight', [1])
+  const anchor1 = await createAnchor()
+  await tGroup.get(0).generate(1)
+  await tGroup.waitForSync()
```
canonbrother, line 81: minor impr
1. remove `waitForSync()` since the test only use 1 node 
2. test create 2 anchors within a block which just
```ts
await createAnchor()
await createAnchor()
await generate(1) // 2 blocks in 1 block
```
why do we need this because the current query would not handle this data..

siradji: @canonbrother i am a little bit confused here. Could you provide more clarification?

canonbrother: create 2 anchors within a block
```ts
await createAnchor()
await createAnchor()
await generate(1) // 2 anchors in 1 block
```

canonbrother: just updated the fixture which contain 2 anchors within a btc block
the query is needed to handle this scenario

fuxingloh:
@siradji please resolve the comments and fix the test

canonbrother:
@siradji 
we need solve the scenario on commit `0d81172`

siradji:
> @siradji please resolve the comments and fix the test

@fuxingloh  i will be fixing it today. Found out a new issue with the query.

siradji:
Initially, i thought that using `startBtcHeight` will be ideal for pagination but due to how the anchors are listed (from top to genesis) that would be impossible. The only option left is to use `maxBtcHeight` which worked out fine till @canonbrother pointed out a [scenario](https://github.com/DeFiCh/whale/commit/0d81172c82ab11a0b8906e079d6e1099e0639520) where there is more than one anchor in a block.

PR #732, opened 2022-01-27
chore(database): add cbor for value encoding to reduce overall storage footprint

fuxingloh:

/kind chore

http://cbor.io/ - a drop-in replacement for JSON

Reducing the overall storage footprint reduces the overall size of whale database snapshots allowing whale-api to startup faster. About 25% reduction based on the written test.

Related to https://github.com/DeFiCh/jellyfish/issues/979


fuxingloh:
Problem with null & undefined handling.

> https://github.com/kawanet/msgpack-lite

> Note that both null and undefined are mapped to nil 0xC1 type. This means undefined value will be upgraded to null in other words.

chee-chyuan (approved):

src/module.database/provider.level/cbor.encoding.ts
```diff
@@ -1,19 +1,19 @@
...
+import cbor from 'cbor'
 
 /**
  * Faster & Smaller Value Codec for level.database.ts
  *
  * @see https://msgpack.org/
  * @see https://github.com/kawanet/msgpack-lite
  */
-export const MsgpackEncoding: CodecEncoder = {
+export const CborEncoding: CodecEncoder = {
```
canonbrother, line 11: we need update the desc above as well though

fuxingloh: It's updated with recent PR

fuxingloh: https://github.com/DeFiCh/whale/pull/732/commits/f8f3553e6b943735ec4b37b86767906104b4296b

canonbrother (approved):

monstrobishi (approved):
test looks good

PR #743, opened 2022-01-31
chore(indexer): removed await within test due to race condition

fuxingloh:

/kind chore

As suggested by @monstrobishi in #733 to remove await within tests that were added due to race condition.

### DIFF
--- a/src/e2e.module.ts
+++ b/src/e2e.module.ts
@@ -44,10 +44,8 @@ export async function stopTestingApp (container: MasterNodeRegTestContainer | Te
     await indexer.stop()
     await app.close()
   } finally {
-    await new Promise((resolve) => {
-      // Wait 2000ms between indexer cycle time to prevent database error
-      setTimeout(_ => resolve(0), 500)
-    })
+    // Wait 500ms between indexer cycle time to prevent database error
+    await new Promise((resolve) => setTimeout(resolve, 500))
 
     if (container instanceof MasterNodeRegTestContainer) {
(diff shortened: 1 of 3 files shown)


src/e2e.module.ts
```diff
@@ -44,10 +44,8 @@ export async function stopTestingApp (container: MasterNodeRegTestContainer | Te
     await indexer.stop()
     await app.close()
   } finally {
-    await new Promise((resolve) => {
-      // Wait 2000ms between indexer cycle time to prevent database error
-      setTimeout(_ => resolve(0), 500)
-    })
+    // Wait 500ms between indexer cycle time to prevent database error
+    await new Promise((resolve) => setTimeout(resolve, 500))
```
monstrobishi, line 48: I see the comment fix and refactor, but we should still wait 500 ms here?

monstrobishi (approved):

src/e2e.module.ts
fuxingloh: Hmm, we might be able to remove this too with https://github.com/DeFiCh/whale/pull/653

monstrobishi:
Seems like the active price test failure is due to the test executing faster, or some similar difference in timing. That test has proven to be flakey, will add a ticket to investigate at some stage.

monstrobishi:
https://github.com/DeFiCh/whale/issues/747

monstrobishi (approved):

canonbrother (approved):

PR #750, opened 2022-02-03
feat(whale-api): get account history

### LINKED ISSUE #718: Get Account History: `/address/${address}/history/${height}/${txn-order}`

Currently, we have List AccountHistory via `/address/${address}/history`. However, we need an endpoint to get single account history.

Endpoint: `/address/${address}/history/${height}/${txn-order}`
Return: `AddressHistory`

/triage accepted
/priority important-soon
/area module-api

canonbrother:

/kind feature

Fixes #718


src/module.api/address.controller.ts
```diff
@@ -31,6 +31,21 @@ export class AddressController {
...
 
+  @Get('/history/:txidn')
+  async getAccountHistory (
+    @Param('address') address: string,
+      @Param('txidn') txidn: string
+  ): Promise<AddressHistory> {
+    const list = await this.listAccountHistory(address, { size: 100 })
+    const txid = txidn.split('-')[0]
+    const n = Number(txidn.split('-')[1])
+    const found = list.data.find(each => each.txid === txid && each.txn === n)
```
canonbrother, line 42: @fuxingloh  seems doesn't need `height` on API :thinking:

src/module.api/address.controller.ts
```diff
@@ -31,6 +31,21 @@ export class AddressController {
   ) {
   }
 
+  @Get('/history/:txidn')
+  async getAccountHistory (
+    @Param('address') address: string,
+      @Param('txidn') txidn: string
+  ): Promise<AddressHistory> {
+    const list = await this.listAccountHistory(address, { size: 100 })
```
fuxingloh, line 39: > We can't do this as it is a scan operation. We need log(n) operation, hence we need to specify:
> 
> 1. size = 1
> 2. maxBlockHeight = `height`
> 3. depth = 1
> 4. no_rewards = true
> 5. ...

Damn it, as I am writing I also just realize there isn't a way to specify txid nor txn-no.

fuxingloh: For this issue, I open an issue on DeFiCh/jellyfish and DeFiCh/ain as we need a new endpoint 

https://github.com/DeFiCh/jellyfish/issues/1025

fuxingloh: We need to wait for that first.

PR #794, opened 2022-02-21
Add volume and commission to `poolpair.controller`

monstrobishi:

/kind feature

Fixes #


monstrobishi:
Note removal of 30d multi-poolpair test due to regression issue. This volume data is assumed to be valid for 30d volume only.

Issue #809, opened 2022-02-28
`DEX` - path finding API


thedoublejay:
#### What would you like to be added:
Currently on the light wallet, we have a path finding algorithm that is used for two things:

1. Display all possible paths (for swapping)
2. Calculate the DEX price based on the path available

With this calculation done on the client side, it is showing some noticeable performance issue due to the heavy calculations. We would like to move both of these two the whale side, so we just need to call the API on the client side.

- [ ] API to get all possible paths for one token. E.g, From `USDT` then return all tokens that can be swapped from it.

![image](https://user-images.githubusercontent.com/25013382/155929856-6c1954ec-dbed-4bc1-b66f-47b658e9b01c.png)

- [ ] DEX price for a specific path, (e.g USDT to DFI, return best price). Will use this also to calculate the portfolio dollar value on the Balances page.
![image](https://user-images.githubusercontent.com/25013382/155929915-b52a4f1a-7c3b-458b-974d-9587b74d5003.png)

- [ ] Optional - Return how much is the estimated token to be received. Ideally should be able to return an array based on the token list. E.g, Passing 1 BTC will return [200 DFI, 30000 USDT, 100 ETH...]

Linked https://github.com/DeFiCh/jellyfish/issues/1065#issuecomment-1053899796

PR #822, opened 2022-03-03
feat(swap-pathfinding): get all paths between tokens

eli-lim:
/kind feature

Implements `/poolpairs/paths/from/1/to/4` endpoint to retrieve all possible swap paths between two tokens
Example response:sadsa
```
                                 ┌─────┐
               ┌────────┐        │     │
               │        ├─ J/K ──┤ K   │
   ┌── I/J ────┤        │        └─┬───┘
   │           │   J    │          │
   │           │        │
┌──┴──┐        └───┬────┘         L/K
│     │            │
│  I  │            │               │
│     │            │           ┌───┴───┐
└─────┘            │           │       │
                   └─ J/L ─────┤   L   │
                               │       │
                               └───────┘
```
```json
{
  "fromToken": {
    "id": "15",
    "symbol": "I"
  },
  "toToken": {
    "id": "17",
    "symbol": "K"
  },
  "paths": [
    [
      { "poolPairId": "20", "symbol": "I-J" },
      { "poolPairId": "22", "symbol": "J-L" },
      { "poolPairId": "23", "symbol": "L-K" }
    ],
    [
      { "poolPairId": "20", "symbol": "I-J" },
      { "poolPairId": "21", "symbol": "J-K" }
    ]
  ]
}
```

Fixes part of #809 

#### Additional comments?:

- No database mapper is required since everything required is derived from


src/module.api/poolpair.service.ts
```diff
@@ -314,3 +319,141 @@ export class PoolPairService {
...
+    result.paths = await this.computePathsBetweenTokens(fromTokenId, toTokenId)
+    return result
+  }
+
+  /**
+   * Performs graph traversal to compute all possible paths between two tokens.
+   * Must be able to handle cycles.
+   * @param {number} fromTokenId
+   * @param {number} toTokenId
+   * @private
```
canonbrother, line 380: ??
```suggestion
```

canonbrother: oh.. i thought that's mistake but that's not.. perhaps add a `return` info??

src/module.api/poolpair.controller.ts
```diff
@@ -113,6 +114,14 @@ export class PoolPairController {
       return `${item.bucket}`
     })
   }
+
+  @Get('/paths/from/:fromToken/to/:toToken')
```
canonbrother, line 118:
```suggestion
  @Get('/paths/from/:fromTokenId/to/:toTokenId')
```

src/module.api/poolpair.service.ts
eli-lim: should tweak my intellij docs generation settings.. 😅
thanks!

canonbrother (approved):

ivan-zynesis:
Nice usage of lib. Wonder will it be better to use Dijstra with weightage (DEX rate) straight.

eli-lim:
> Nice usage of lib. Wonder will it be better to use Dijstra with weightage (DEX rate) straight.
Yea it's a great lib

@thedoublejay do we have a use-case for all possible paths? or do we only ever want the best?
alternatively, we can support both - i'll create another endpoint to serve _the best path only_

thedoublejay:
@eli-lim Yep! We use it to display all the possible tokens to be swapped to. Example, if I select USDT, I need to display all possible tokens that can be swapped. For now, all tokens are displayed. But in the future we might have an unreachable pair

eli-lim:
@thedoublejay ah okay, but "all possible/swappable tokens" probably should show something like this:

```json
{
  "fromToken": {
    "id": "1",
    "symbol": "A"
  },
  "swappableTokens": [
    { "id": "2", "symbol": "B" },
    { "id": "3", "symbol": "C" },
    { "id": "4", "symbol": "D" }
  ]
}
```
instead of paths of poolPairs.

So in the end we'll have two endpoints:
- get all swappable tokens for a given token
- get the best path from one token to another

Hence we won't need "all possible paths" - we only need the best

thedoublejay:
@eli-lim Ok great! should be ok on our end

eli-lim:
> Nice usage of lib. Wonder will it be better to use Dijstra with weightage (DEX rate) straight.

Tried, but don't think it's possible due to:
- In order to compute the max return of a given swap path, the edge weights need to be multiplied together instead of added. There's a trick to do this with logarithms - https://www.cs.virginia.edu/~jh2jf/courses/fall2019/cs4102/lectures/robbie/l21.pdf
- However, we're trying to find the MAXIMUM return (djikstra finds the minimum). Negating the values won't work as Djikstra doesn't work with negative values.

Hence i've opted for brute-force for now - get all the possible paths and return the best.

We can invest time to improve this in the future if necessary.

canonbrother (dismissed):

src/module.api/poolpair.service.ts
```diff
@@ -425,3 +429,194 @@ function findPoolSwapFromTo (history: AccountHistory, from: boolean): PoolSwapFr
...
+export class PoolSwapPathFindingService {
+  tokenGraph: UndirectedGraph = new UndirectedGraph()
+
+  constructor (
+    protected readonly poolPairTokenMapper: PoolPairTokenMapper,
+    protected readonly deFiDCache: DeFiDCache
+  ) {
+  }
+
+  @Interval(10_000) // 10s
```
fuxingloh, line 443:
```suggestion
  @Interval(120_000) // 10s
```

In the real world, the PoolPair is mostly constant, maybe we can change it to a larger number? 🤔

src/module.api/poolpair.service.ts
```diff
@@ -425,3 +429,194 @@ function findPoolSwapFromTo (history: AccountHistory, from: boolean): PoolSwapFr
...
+  }
+
+  @Interval(120_000) // 120s
+  async syncTokenGraph (): Promise<void> {
+    const poolPairTokens = await this.poolPairTokenMapper.list(200)
+    await this.addTokensAndConnectionsToGraph(poolPairTokens)
+  }
+
+  async getBestPath (fromTokenId: string, toTokenId: string): Promise<BestSwapPathResult> {
+    const { paths } = await this.getAllSwapPaths(fromTokenId, toTokenId)
```
ivan-zynesis, line 450: Can add `if (paths.length === 0）attemptAddTokensAndConnectionsToGraph()`

eli-lim: That's already handled in `getAllSwapPaths`

fuxingloh:
What are the performance implications of deploying this?

fuxingloh:
@pierregee need your review on this too

eli-lim:
> What are the performance implications of deploying this?

- Memory: not much requiered since the in-memory graph only stores tokens and their links based on the existing poolpairs
- Query performance: right now there are about 30 poolpairs - most of the tokens are bridged via dfi / dusd. hence I don't expect the paths traversed / returned to be long / complex.

pierregee (approved):
Great find on the lib and test coverage! 👍  Looks good to me and it'll cover our use case in light wallet

fuxingloh (approved):
nice

PR #823, opened 2022-03-03
feat(api/swaps/aggregated): include USD value for aggregated swaps

fuxingloh:

/kind feature

Resolve USD value of AggregatedSwap at current dex prices on controller.

For graphing on https://deploy-preview-1008--defi-scan.netlify.app/dex/5

### DIFF
--- a/packages/whale-api-client/__tests__/api/poolpairs.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpairs.test.ts
@@ -598,19 +598,8 @@ describe('poolswap aggregated', () => {
     expect([...dayAggregated]).toStrictEqual([
       {
         aggregated: {
-          amounts: { 2: '9.50000000' }
-        },
-        block: expect.any(Object),
-        bucket: expect.any(Number),
-        id: expect.any(String),
-        key:
-        '10-86400'
-      },
-      {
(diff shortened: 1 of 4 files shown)


ivan-zynesis (approved):
Looks good.

Just that history asset amount evaluations are with current USD rate sounds weird to me. (very subjective to context)

joeldavidw:
> Looks good.
> 
> Just that history asset amount evaluations are with current USD rate sounds weird to me. (very subjective to context)

Yea same sentiments. This data is required for graphs on scan but with only the latest USD rate the graphs would be inaccurate.

fuxingloh:
TBH, I follow the same sentiments as well. As dex event prices aren't streamed out in real-time from the node and there isn't a reliable way to capture those data for indexing. 

This is the best we could do (for now) but the best is just not enough as pointed out from the above 2 comments. Let's keep the PR as it is and see what else we could do.

fuxingloh:
I am going to merge this in, but the consumers of this endpoint should use it sparingly...

fuxingloh:
> I am going to merge this in, but the consumers of this endpoint should use it sparingly...

We will plan another API integration for historical dex prices, since we need that too

PR #844, opened 2022-03-11
Added buy/sell indicator for each swap entry

nichellekoh:

/kind feature

Add swap sell/buy indicator 

Fixes #

### DIFF
--- a/src/module.api/poolpair.service.ts
+++ b/src/module.api/poolpair.service.ts
@@ -219,7 +219,7 @@ export class PoolPairService {
     return value
   }
 
-  public async findSwapFromTo (height: number, txid: string, txno: number): Promise<{ from?: PoolSwapFromToData, to?: PoolSwapFromToData } | undefined> {
+  public async findSwapFromTo (height: number, txid: string, txno: number): Promise<{ from?: PoolSwapFromToData, to?: PoolSwapFromToData, isSell: boolean } | undefined> {
     const vouts = await this.voutMapper.query(txid, 1)
     const dftx = findPoolSwapDfTx(vouts)
     if (dftx === undefined) {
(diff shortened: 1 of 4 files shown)


src/module.api/poolpair.service.ts
```diff
@@ -239,13 +239,22 @@ export class PoolPairService {
       return undefined
     }
 
+    const poolPairInfo = await this.deFiDCache.getPoolPairInfo(id)
+    if (poolPairInfo === undefined) {
+      return undefined
+    }
+    const [,symbolB] = poolPairInfo.symbol.split('-')
```
fuxingloh, line 246: > Currently, there are some limitations from the upstream limiting the ability to track swaps between composite. For example a swap of ETH-DFI-DUSD through ETH-DFI, DFI-DUSD pairs. Light clients (ocean), don’t know what’s happening as it’s calculated automatically on `defid` side. Hence we are indexing `from-to` of composite swaps not relative to the PoolPair path but in absolute to the entire swap. 

> This implementation is a compromised approach that we had to take to have some visibility in each pair. In the long run, the upstream team is working on exposing the native events directly for us to index separately without having to take this compromised path.

You won't be able to catch a composite path swap with this logic:

e.g. [A-DFI, B-DFI, B-C]

Indexed Composite Swap: `A-C`
Actual Composite Path: `A-DFI, DFI-B, B-C`
From `A` to `C`, it was routed through `B-DFI` pair.

In `B-DFI`, the path it took converts `DFI` to `B`, hence a `buy` order. However, the current logic will return a `sell` order.

nichellekoh: Thanks for breaking the logic down, will take this into consideration and modify the logic. This PR should be in draft mode for now, as it is currently still pending.

fuxingloh: image

You can convert to draft on the top right hand side

fuxingloh: This affects https://github.com/DeFiCh/whale/pull/847 implementation as well

PR #847, opened 2022-03-14
Add to listPoolSwapsVerbose to return displaySymbol of swap

nichellekoh:

/kind feature

Return the swap displaySymbols with the API to allow any UI to extract the right symbol for swap

Fixes #

### DIFF
--- a/packages/whale-api-client/__tests__/api/poolpairs.test.ts
+++ b/packages/whale-api-client/__tests__/api/poolpairs.test.ts
@@ -474,16 +474,14 @@ describe('poolswap', () => {
         from: {
           address: expect.any(String),
           symbol: 'A',
-          amount: '50.00000000'
+          amount: '50.00000000',
+          displaySymbol: 'dA'
         },
         to: {
           address: expect.any(String),
           amount: '45.71428571',
-          symbol: 'DFI'
-        },
-        displaySymbols: {
-          displaySymbolA: 'dA',
-          displaySymbolB: 'DFI'
+          symbol: 'DFI',
+          displaySymbol: 'DFI'
         }
       },
       {
@@ -503,16 +501,14 @@ describe('poolswap', () => {
         from: {
           address: expect.any(String),
           symbol: 'A',
-          amount: '25.00000000'
+          amount: '25.00000000',
+          displaySymbol: 'dA'
         },
         to: {
           address: expect.any(String),
           amount: '39.99999999',
-          symbol: 'DFI'
-        },
-        displaySymbols: {
-          displaySymbolA: 'dA',
-          displaySymbolB: 'DFI'
+          symbol: 'DFI',
+          displaySymbol: 'DFI'
         }
       }
     ])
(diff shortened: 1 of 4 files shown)


packages/whale-api-client/__tests__/api/poolpairs.test.ts
```diff
@@ -432,6 +432,10 @@ describe('poolswap', () => {
           address: expect.any(String),
           amount: '45.71428571',
           symbol: 'DFI'
+        },
+        displaySymbols: {
```
fuxingloh, line 436: More structurally accurate if you put them in `to` and `from`, since A,B is relative

nichellekoh: Have refactored as per suggested! 👍🏼

nichellekoh: To clarify my understanding, so in the case where
Indexed Composite Swap: A-C
Actual Composite Path: A-DFI, DFI-B, B-C

For the DFI-B swap pair page, the icons indicating the from-to swap should show A-C?

Screenshot 2022-03-14 at 12 27 19 PM

Therefore, in an actual example above should display LTC -> dBTC in the pool pair table?

fuxingloh: Yea

fuxingloh: I updated PR https://github.com/DeFiCh/whale/pull/848 with that example in test for your reference

nichellekoh: Updated logic, and passed the composite swap test case added

src/module.api/poolpair.service.ts
```diff
@@ -411,15 +417,17 @@ function findPoolSwapFromTo (history: AccountHistory | undefined, from: boolean)
       return {
         address: history.owner,
         amount: new BigNumber(value).absoluteValue().toFixed(8),
-        symbol: symbol
+        symbol: symbol,
+        displaySymbol: displaySymbol
```
ivan-zynesis, line 421: Can `parseDisplaySymbol()` here instead of taking in argument.

ivan-zynesis (changes requested):

src/module.api/poolpair.service.ts
nichellekoh: Updated in this [commit](https://github.com/DeFiCh/whale/pull/847/commits/4c4c232e07754fedbfc01d499a1c666ddf19601a)

nichellekoh: Do ignore my previous deleted comment. 
`parseDisplaySymbol()` takes in a `TokenInfo` param which is not available within this function. Is there another way to do so?

fuxingloh: Yeah, it's kinda awkward hmm. But don't think there is a way around it unless we refactor the entire function.

fuxingloh:
@nichellekoh there is a conflict

fuxingloh (approved):

PR #848, opened 2022-03-14
chore(poolpair.controller): allow `to` to be optional when history cannot be resolved

fuxingloh:

/kind chore

- This PR allows `to` to be optional when history cannot be resolved.
- Also add a `CompositeSwap` transaction to `poolpairs.test.ts` so that all the correct scenarios are tested.


fuxingloh:
@nichellekoh I added CompositeSwap to the test as an example so that your future PR can test it as well.

nichellekoh (approved):

PR #849, opened 2022-03-14
feat(controller): token swap pathfinding getallswappabletokens

eli-lim:
#### What kind of PR is this?:
/kind feature

#### What this PR does / why we need it:
- Implements `/poolpairs/paths/swappable/1` endpoint to retrieve all swappable tokens given a token
- To merge after #822 

Returns response:
`poolswap/paths/swappable/1`
```javascript
{
  fromToken: { id: '1', symbol: 'A' },
  swappableTokens: [
    { id: '7', symbol: 'G' },
    { id: '0', symbol: 'DFI' },
    { id: '24', symbol: 'USDT' },
    { id: '6', symbol: 'F' },
    { id: '5', symbol: 'E' },
    { id: '4', symbol: 'D' },
    { id: '3', symbol: 'C' },
    { id: '2', symbol: 'B' }
  ]
}
```

#### Which issue(s) does this PR fixes?:
Fixes part of #809

PR #854, opened 2022-03-15
feat(swap-pathfinding): glue api client with swap path finding endpoints

eli-lim:
/kind feature

Exposes functionality via whale-api-client that was added in:
- #822 
- #849

Fixes part of https://github.com/DeFiCh/whale/issues/809

#### Additional comments?:
- changed ParseIntPipe to StringIsInteger validation pipe so a String value is passed to the controller instead of a Number (all places using ParseIntPipe but expecting a string are lulled into a false sense of security)

### DIFF
--- a/packages/whale-api-client/src/api/poolpairs.ts
+++ b/packages/whale-api-client/src/api/poolpairs.ts
@@ -67,17 +67,30 @@ export class PoolPairs {
   }
 
   /**
-   * List all swappable tokens for a given token
+   * Get all swappable tokens for a given token
    * @param {string} tokenId
+   * @return {Promise<AllSwappableTokensResult>}
    */
   async getSwappableTokens (tokenId: string): Promise<AllSwappableTokensResult> {
     return await this.client.requestData('GET', `poolpairs/paths/swappable/${tokenId}`)
   }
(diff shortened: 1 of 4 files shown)


fuxingloh (approved):

packages/whale-api-client/src/api/poolpairs.ts
```diff
@@ -65,6 +65,22 @@ export class PoolPairs {
...
+
+  /**
+   * List all swappable tokens for a given token
+   * @param {string} tokenId
+   */
+  async getSwappableTokens (tokenId: string): Promise<AllSwappableTokensResult> {
+    return await this.client.requestData('GET', `poolpairs/paths/swappable/${tokenId}`)
+  }
+
+  async getBestPath (fromTokenId: string, toTokenId: string): Promise<SwapPathsResult> {
```
thedoublejay, line 77: Can we add docs here and for getAllPaths?

eli-lim: done

thedoublejay (approved):

fuxingloh (approved):

PR #863, opened 2022-03-17
fix(module.api): align stats.burn with stats.supply.burn

### LINKED ISSUE #845: burn information in `/stats` & `/stats/supply` is not aligned
#### What happened:

The burn information for endpoint `/stats` & `/stats/supply` is not aligned. As both endpoints use different cache windows and ways to derivate the burn information there is minor drift in the burn amount.

#### What you expected to happen:

We should align `/stats` burn info to use `/stats/supply` burn information.

canonbrother:

/kind fix

Fixes #845

### DIFF
--- a/src/module.api/stats.controller.ts
+++ b/src/module.api/stats.controller.ts
@@ -28,13 +28,17 @@ export class StatsController {
   @Get()
   async get (): Promise<StatsData> {
     const block = requireValue(await this.blockMapper.getHighest(), 'block')
-
+    const burned = await this.cachedGet('burned', this.getBurned.bind(this), 1806)
+    const burnedTotal = await this.cachedGet('Controller.supply.getBurnedTotal', this.getBurnedTotal.bind(this), 1806)
     return {
       count: {
         ...await this.cachedGet('count', this.getCount.bind(this), 1801),
         blocks: block.height
       },
-      burned: await this.cachedGet('burned', this.getBurned.bind(this), 1700),
+      burned: {
+        ...burned,
+        total: burnedTotal.toNumber()
+      },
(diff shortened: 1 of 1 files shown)


src/module.api/stats.controller.ts
```diff
@@ -28,13 +28,17 @@ export class StatsController {
...
     return {
       count: {
         ...await this.cachedGet('count', this.getCount.bind(this), 1801),
         blocks: block.height
       },
-      burned: await this.cachedGet('burned', this.getBurned.bind(this), 1700),
+      burned: {
+        ...burned,
+        total: burnedTotal.toNumber()
+      },
```
canonbrother, line 41: Checked scan that only 1 space is using `stats.burned.total`
https://github.com/DeFiCh/scan/blob/de27e32fa9b51f048ddc46388ab2200e04478596/src/components/index/IndexHeader.tsx#L59

only overwrite the burnedTotal to prevent regression fix on scan
remain the rpc.getburninfo easier for downstream

packages/whale-api-client/__tests__/api/stats.test.ts
```diff
@@ -90,7 +90,7 @@ describe('stats', () => {
         emission: 7323.58,
         fee: 4,
         payback: 0,
-        total: 7327.58
+        total: 7323.58
```
fuxingloh, line 93: Hmm, I think we should probably introduce fee burning calculation into total? for `/stats/supply` as well. Seems like it's missing, my assumption was this was included as part of the Account layer calculation but I realized it's not.

canonbrother: just aware the "missing" is from "fee" :sweat_smile:

fuxingloh:
Regardless, let's merge

PR #869, opened 2022-03-21
add test cases

### LINKED ISSUE #739: Populate rewardPct for stock LPs
The percentage distribution of block rewards for stock (ie. DUSD-based) liquidity pools. This should be recorded in the _rewardPct_ attribute under the poolpairs access point.

#### Why is this needed:
The attributes of these assets should mirror the DFI-based LPs (e.g. ETH-DFI). There is no clear way under the API to extract the reward distribution of the stock LPs, whereas there is with DFI-based pools. Given the former are subject to change, it is very useful to have the correct distribution available.

ivan-zynesis:

/kind feature

Include reward split % for DUSD pairs

Fixes #739

### DIFF
--- a/src/module.api/poolpair.service.ts
+++ b/src/module.api/poolpair.service.ts
@@ -60,7 +60,7 @@ export class PoolPairService {
       if (Object.values(result).length > 0) {
         return Object.values(result)[0]
       }
-    } catch (err) {
+    } catch (err: any) {
       if (err?.payload?.message !== 'Pool not found') {
         throw err
       }
@@ -71,7 +71,7 @@ export class PoolPairService {
       if (Object.values(result).length > 0) {
         return Object.values(result)[0]
       }
-    } catch (err) {
+    } catch (err: any) {
       if (err?.payload?.message !== 'Pool not found') {
         throw err
       }
@@ -386,6 +386,23 @@ export class PoolPairService {
       commission: commission.toNumber(),
       total: reward.plus(commission).toNumber()
     }
+  }
+
+  async getRewardPct (id: string, info: PoolPairInfo): Promise<BigNumber> {
+    if (!info.rewardPct.isZero()) {
+      return info.rewardPct
+    }
+
+    const token = await this.deFiDCache.getTokenInfo(info.idTokenA)
+    if (token === undefined) {
+      throw new Error(`Pool ${id} not found`)
+    }
+
+    if (!token.isLoanToken) {
+      return new BigNumber(0)
+    }
+
+    return await this.deFiDCache.getStockLpRewardPct(id)
   }
 }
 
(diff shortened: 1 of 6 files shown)


canonbrother (approved):

ivan-zynesis:
@fuxingloh this is merge ready and should be quite minimal impact to any downstream

src/module.api/poolpair.service.ts
```diff
@@ -60,7 +60,7 @@ export class PoolPairService {
       if (Object.values(result).length > 0) {
         return Object.values(result)[0]
       }
-    } catch (err) {
+    } catch (err: any) {
       if (err?.payload?.message !== 'Pool not found') {
```
kodemon, line 64:
```suggestion
    } catch (err) {
      if (err instanceof RpcApiError && err?.payload?.message !== 'Pool not found') {
```
We can potentially use instanceof here to avoid casting errors to any

src/module.api/poolpair.service.ts
```diff
@@ -71,7 +71,7 @@ export class PoolPairService {
...
```
kodemon, line 75:
```suggestion
    } catch (err) {
      if (err instanceof RpcApiError && err?.payload?.message !== 'Pool not found') {
```

src/module.api/poolpair.service.ts
kodemon: Note that this requires `RpcApiError` to be imported so will need manual commits I think 🤔

ivan-zynesis: Quite reluctant to make changes here, since it is a generic issue and all over the codebase. I think we should open an issue to clean it up if we have to.

kodemon: I am wondering if we want to set the typescript rules to use the [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) until we formally address this so we don't have to cast back to any for our try/catch.

But that could also be part of a separate PR.

ivan-zynesis: Omg, I didn't actually look deep into this. JS 😅 
image

fuxingloh:
We don't need to manually calculate it anymore since `listpoolpairs` exposes it via `rewardLoanPct` and we can just expose that now.

fuxingloh:
> We don't need to manually calculate it anymore since `listpoolpairs` exposes it via `rewardLoanPct` and we can just expose that now.

I guess this means we can also simplify the code and remove the manual calculation all together and just get the data from listPoolPair directly.

fuxingloh:
We're finishing up https://github.com/JellyfishSDK/jellyfish/issues/580

This PR needs to be migrated to https://github.com/JellyfishSDK/jellyfish/tree/main/apps/whale-api

Since you're busy currently, I will re-assign this to someone else.

PR #870, opened 2022-03-21
feat(poolpairs/prices): add poolpairs/dexprices endpoint

eli-lim:
/kind feature

- Adds the `poolpairs/dexprices?denomination=XXX` endpoint
- Deliberately named `dexprices` to avoid confusion with `ocean.defichain.com/v0/mainnet/prices`

Fixes part of #2318

#### Additional comments:
- Moved pathfinding service into its own file
- Added logic in pathfinding service to disallow `getPath({fromToken: 'DFI', toToken: 'DFI})` (from and to are the same token)  
- Fixes all '0.00000000' to '0' for consistency and less data over the wire


src/module.api/poolswap.pathfinding.service.ts
```diff
@@ -0,0 +1,233 @@
...
+        bestPath = path
+      }
+    }
+    return {
+      fromToken: fromToken,
+      toToken: toToken,
+      bestPath: bestPath,
+      estimatedReturn: bestReturn.eq(-1)
+        ? '0'
+        : bestReturn.toFixed(8) // denoted in toToken
```
ivan-zynesis, line 64:
```suggestion
        : bestReturn.dp(8).toFixed() // denoted in toToken
```
I think will be cleaner and more consistent.

src/module.api/poolpair.prices.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  private readonly logger: Logger = new Logger(PoolSwapPathFindingService.name)
+
+  constructor (
+    private readonly poolSwapPathfindingService: PoolSwapPathFindingService,
+    private readonly defidCache: DeFiDCache,
+    private readonly tokenMapper: TokenMapper
+  ) {
+  }
+
+  async listDexPrices (denominationSymbol: string): Promise<DexPricesResult> {
```
ivan-zynesis, line 20: Maybe should cache this, for duration a minute or so. `getBestPath()` for all tokens are quite heavy and esp when the deno token is not DFI nor DUSD, the paths are long.

src/module.api/poolswap.pathfinding.service.ts
eli-lim: Interesting - I see .toFixed(8) being used more across whale and jellyfish though

src/module.api/poolpair.prices.service.ts
eli-lim: yea agree. will do that, thanks! :)

eli-lim:
@pierregee @thedoublejay made the endpoint query param based on `symbol` instead of `token id` for ease of use. Is there a preference?

src/module.api/poolswap.pathfinding.service.ts
ivan-zynesis: Or make the previous line `? '0.00000000'` works too. I guess it is depend on the result consumer, for API, less data can be considered better.

eli-lim: Oh my bad, i didn't realise `.dp(8).toFixed()` truncates the leading zeroes

I fixed it to make it all consistently return '0' another way though

thedoublejay:
@eli-lim looks ok to me! Just want to check, is it `symbol` or `displaySymbol`?

eli-lim:
@thedoublejay symbol!

ivan-zynesis (approved):

src/module.api/poolpair.prices.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  /**
+   * Helper to get all tokens in a map indexed by their symbol for quick look-ups
+   * @private
+   * @return {Promise<TokensBySymbol>}
+   */
+  private async getAllTokens (): Promise<TokensBySymbol> {
+    const tokens = await this.tokenMapper.query(200)
```
canonbrother, line 81: i'm afraid we will come back here again because its token.. maybe??
```suggestion
    const tokens = await this.tokenMapper.query(Number.MAX_SAFE_INTEGER)
```
or for loop the `mapper.query`...

src/module.api/poolpair.controller.ts
```diff
@@ -183,6 +186,13 @@ export class PoolPairController {
   ): Promise<BestSwapPathResult> {
     return await this.poolSwapPathService.getBestPath(fromTokenId, toTokenId)
   }
+
+  @Get('/dexprices')
+  async listPrices (
+    @Query('denomination') denomination: string
+  ): Promise<DexPricesResult> {
+    return await this.poolPairPricesService.listDexPrices(denomination)
```
canonbrother, line 194: hmm... better give them a same name

src/module.api/poolpair.controller.ts
```diff
@@ -183,6 +186,13 @@ export class PoolPairController {
   ): Promise<BestSwapPathResult> {
     return await this.poolSwapPathService.getBestPath(fromTokenId, toTokenId)
   }
+
+  @Get('/dexprices')
+  async listPrices (
```
canonbrother, line 191:
```suggestion
  async listDexPrices (
```
better give them a same..

canonbrother (approved):

ivan-zynesis (approved):

PR #876, opened 2022-03-23
feat(module.api): `getAccountHistory` api

### LINKED ISSUE #718: Get Account History: `/address/${address}/history/${height}/${txn-order}`

canonbrother:

/kind feature

Fixes #718

#### Additional comments?:
Current `getAccountHistory` doesn't support to get wallet kind transactions, eg: `sent`, `receive`, `blockReward`, if intended, will return undefined


packages/whale-api-client/__tests__/api/account.history.test.ts
```diff
@@ -103,66 +103,104 @@ afterAll(async () => {
...
+      expect(acc?.owner).toStrictEqual(h.owner)
+      expect(acc?.txid).toStrictEqual(h.txid)
+      expect(acc?.txn).toStrictEqual(h.txn)
+    }
+  })
+
+  it('should get undefined as getting unsupport tx type - sent, received, blockReward', async () => {
+    const history = await client.address.listAccountHistory(colAddr, 30)
+    for (const h of history) {
+      if (['sent', 'receive', 'blockReward'].includes(h.type)) {
```
ivan-zynesis, line 200: Is the test actually setup with rewards in history?

```
const rewards = history.filter(h => h.type === 'blockReward')
expect(rewards.length).toBeGreaterThan(0)
```

ivan-zynesis (commented):
It is actually quite annoying that upstream `listaccounthistory` return `sent` and `receive` (utxo transaction) too, unfortunately doesn't seems will be fixed anytime soon.

packages/whale-api-client/__tests__/api/account.history.test.ts
canonbrother: checked `RegTestFoundationKeys[0].operator.address`'s `blockReward` history

ivan-zynesis (approved):

src/module.api/address.controller.e2e.ts
```diff
@@ -209,6 +210,146 @@ describe('listAccountHistory', () => {
...
+      }
+      const acc = await controller.getAccountHistory(poolAddr, h.block.height, h.txn)
+      expect(acc?.owner).toStrictEqual(h.owner)
+      expect(acc?.txid).toStrictEqual(h.txid)
+      expect(acc?.txn).toStrictEqual(h.txn)
+    }
+  })
+
+  it('should get undefined as invalid height', async () => {
+    const acc = await controller.getAccountHistory(await container.getNewAddress(), Number(`${'0'.repeat(64)}`), 1)
```
ivan-zynesis, line 330: The number input here should be still zero, can test with NaN.

Nestjs doesn't seems to built in with positive number only pipe, maybe can test with negative number too.

ivan-zynesis (approved):

fuxingloh (approved):

PR #903, opened 2022-04-08
fix(poolpair/paths): make defidcache.getpoolpairinfo rely on 1m cached list instead of 10m cache

eli-lim:
/kind fix

- Fixes price updates taking too long due to reliance on 10m TTL caching for poolpairinfo

#### Additional comments?:
- Pagination needs to be followed in order to retrieve the full list - this is made explicit with a `while (true)`

### DIFF
--- a/src/module.api/cache/defid.cache.ts
+++ b/src/module.api/cache/defid.cache.ts
@@ -58,19 +58,44 @@ export class DeFiDCache extends GlobalCache {
   }
 
   async getPoolPairInfo (id: string): Promise<PoolPairInfo | undefined> {
-    const poolPairsById = await this.listPoolPairs()
+    return await this.get<PoolPairInfo>(CachePrefix.POOL_PAIR_INFO, id, this.fetchPoolPairInfo.bind(this))
+  }
+
+  /**
+   * Retrieve poolPair info from cached list of poolPairs as getPoolPair rpc
+   * tends to be more expensive
+   * @param {string} id - id of the poolPair
+   */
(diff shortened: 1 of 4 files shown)


canonbrother (approved):

src/module.api/cache/defid.cache.ts
```diff
@@ -61,6 +61,27 @@ export class DeFiDCache extends GlobalCache {
...
+  async getPoolPairInfoFromPoolPairs (id: string): Promise<PoolPairInfo | undefined> {
+    const poolPairsById = await this.listPoolPairs(60)
+    if (poolPairsById === undefined) {
+      return undefined
+    }
+    return poolPairsById[id]
+  }
+
+  async listPoolPairs (ttlSeconds: number): Promise<PoolPairsResult | undefined> {
+    return await this.get<PoolPairsResult>(CachePrefix.POOL_PAIRS, '*', this.fetchPoolPairs.bind(this),
```
ivan-zynesis, line 78: Thinking out loud, use `CachePrefix.POOL_PAIR_INFO` is also fine.

ivan-zynesis (approved):

src/module.api/cache/defid.cache.ts
eli-lim: 😅 then would need to either:
- update to `this.get<PoolPairsResult | PoolPairInfo>` or
- suppress typescript for the line

ivan-zynesis: That's true, it is caching the entire list instead of one component.

PR #906, opened 2022-04-12
fix(poolpair/dexprices): add denomination symbol to dexprices cache key

eli-lim:
/kind fix

- Fixes the issue of `/dexprices?denomination=SYMBOL` ignoring SYMBOL and returning wrong result
- Issue was that the denomination symbol wasn't being added to `listDexPrices` cache key, so all queries were hitting 'LIST_DEX_PRICES' regardless of the denomination symbol


kodemon (approved):

eli-lim:
@weiyuan95 I think I was assuming the cache was causing problems in the testing environment and jumped the gun and disabled it. But it turns out the only thing causing problems was my brain.

fuxingloh (approved):

PR #909, opened 2022-04-14
feat(module.api): Whitelist `getFutureSwapBlock` RPC

siradji:

Fixes #


src/module.api/rpc.controller.ts
```diff
@@ -35,7 +35,8 @@ export class MethodWhitelist implements PipeTransform {
     'getgov',
     'validateaddress',
     'listcommunitybalances',
-    'getaccounthistory'
+    'getaccounthistory',
+    'getFutureSwapBlock'
```
jingyi2811, line 39: Should this be all lowercase letters?

fuxingloh (approved):

PR #940, opened 2022-05-06
feat(module-indexer, module-api, whale-api-client): Future swap indexer and API

### LINKED ISSUE #939: Indexing FutureSwap on Whale Indexer

Instead of relying on defid for FutureSwap, move the indexing of FutureSwap to whale via and append-only indexing Add/Remove DfTx on whale indexer. 

**keys:**
- partition: address
- sort: height + txid

**notable fields:**
- type: `add` or `remove`
- everything else

**Controller Querying?**

For `AddressController.listFutureSwap`, given that FutureSwap goes into effect at a target block height, the controller should query from target height to minus interval height

/triage accepted
/area module-indexer module-api whale-api-client
/priority important-soon
/assign @canonbrother

canonbrother:

Fixes #939

### DIFF
--- a/packages/whale-api-client/src/api/address.ts
+++ b/packages/whale-api-client/src/api/address.ts
@@ -101,6 +101,19 @@ export class Address {
   async listTransactionUnspent (address: string, size: number = 30, next?: string): Promise<ApiPagedResponse<AddressUnspent>> {
     return await this.client.requestList('GET', `address/${address}/transactions/unspent`, size, next)
   }
+
+  /**
+   * List future swap
+   *
+   * @param {string} address to list future swap
+   * @param {number} height to set range of future swap
+   * @param {string} size to query
+   * @param {string} next token for next slice of FutureSwap
+   * @return {Promise<ApiPagedResponse<FutureSwap>>}
+   */
+  async listFutureSwap (address: string, height: number, size: number = 30, next?: string): Promise<ApiPagedResponse<FutureSwap>> {
+    return await this.client.requestList('GET', `address/${address}/future-swaps/${height}`, size, next)
+  }
 }
 
 /**
@@ -215,6 +228,24 @@ export interface AddressHistory {
   block: {
     height: number
     hash: string
+    time: number
+  }
+}
+
+export interface FutureSwap {
+  id: string
+  key: string
+  sort: string
+  source: {
+    token: number
+    amount: string
+  }
+  destination: number
+  withdraw: boolean
+  block: {
+    hash: string
+    height: number
+    medianTime: number
     time: number
   }
 }
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -189,20 +189,21 @@ export class AddressController {
     })
   }
 
-  @Get('/future-swaps')
+  @Get('/future-swaps/:height')
   async listFutureSwap (
     @Param('address') address: string,
       @Param('height', ParseIntPipe) height: number,
       @Query() query: PaginationQuery
   ): Promise<ApiPagedResponse<any>> {
-    const gt = query.next ?? `${HexEncoder.encodeHeight(height)}-${'0'.repeat(64)}`
-
+    const size = query.size > 30 ? 30 : query.size
     const nextSettleBlock = await this.rpcClient.oracle.getFutureSwapBlock()
-    const lt = `${HexEncoder.encodeHeight(nextSettleBlock)}-${'f'.repeat(64)}`
+    // mapper is sorted DESC
+    const lt = query.next ?? `${HexEncoder.encodeHeight(nextSettleBlock)}-${'f'.repeat(64)}`
+    const gt = `${HexEncoder.encodeHeight(height)}-${'0'.repeat(64)}`
 
-    const list = await this.futureSwapMapper.query(address, query.size, lt, gt)
+    const list = await this.futureSwapMapper.query(address, size, lt, gt)
 
-    return ApiPagedResponse.of(list, query.size, item => {
+    return ApiPagedResponse.of(list, size, item => {
       return item.sort
     })
   }
--- a/src/module.api/address.controller.ts
+++ b/src/module.api/address.controller.ts
@@ -9,6 +9,7 @@ import { PaginationQuery } from '@src/module.api/_core/api.query'
 import { ScriptActivity, ScriptActivityMapper } from '@src/module.model/script.activity'
 import { ScriptAggregation, ScriptAggregationMapper } from '@src/module.model/script.aggregation'
 import { ScriptUnspent, ScriptUnspentMapper } from '@src/module.model/script.unspent'
+import { FutureSwapMapper } from '@src/module.model/future.swap'
 import { DeFiAddress } from '@defichain/jellyfish-address'
 import { NetworkName } from '@defichain/jellyfish-network'
 import { HexEncoder } from '@src/module.model/_hex.encoder'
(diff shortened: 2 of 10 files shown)


nichellekoh (approved):

src/module.model/future.swap.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+export interface FutureSwap extends Model {
+  id: string // -------------| txid
+  key: string // ------------| fromAddress
+  sort: string // -----------| hexEncodedHeight-txid
+  source: {
+    token: number
+    amount: BigNumber
```
jingyi2811, line 53:
```suggestion
    amount: string
```

String is used in Whale because it handles precision better than BigNumber. If you take a look at amount variable in the same folder, amount is declared as string

jingyi2811:
Title should include `module.model`

src/module.indexer/model/dftx/set.future.swap.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+      destination: 2,
+      withdraw: false,
+      block: expect.any(Object)
+    })
+  }
+
+  {
+    const res = await futureSwapMapper.query(fromAddr1, 30)
+    expect(res.length).toStrictEqual(4)
+    expect(res[0].withdraw).toStrictEqual(true)
```
jingyi2811, line 221: You don't need to check withdraw variable again as you have checked it in line
https://github.com/JellyfishSDK/whale/pull/940/files#diff-88f8c7070c8e56bc078ea56647f8fde0182fa2e5b7534d7b5e0feb371e2da8aaR230

Like how you wrote in
https://github.com/JellyfishSDK/whale/pull/940/files#diff-88f8c7070c8e56bc078ea56647f8fde0182fa2e5b7534d7b5e0feb371e2da8aaR203

src/module.api/address.controller.ts
```diff
@@ -186,6 +188,25 @@ export class AddressController {
       return item.sort
     })
   }
+
+  @Get('/future-swaps/:height')
+  async listFutureSwap (
+    @Param('address') address: string,
+      @Param('height', ParseIntPipe) height: number,
+      @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<any>> {
```
jingyi2811, line 197:
```suggestion
  ): Promise<ApiPagedResponse<FutureSwap>> {
```

packages/whale-api-client/src/api/address.ts
```diff
@@ -101,6 +101,19 @@ export class Address {
   async listTransactionUnspent (address: string, size: number = 30, next?: string): Promise<ApiPagedResponse<AddressUnspent>> {
     return await this.client.requestList('GET', `address/${address}/transactions/unspent`, size, next)
   }
+
+  /**
+   * List future swap
+   *
+   * @param {string} address to list future swap
+   * @param {number} height to set range of future swap
+   * @param {string} size to query
```
jingyi2811, line 110: Doc should tell the reader there are 
1) Default value 30
2) Next variable is an optional variable

jingyi2811:
Quite good. Some minor stuff

jingyi2811 (approved):

nichellekoh (dismissed):

cwkang1998 (dismissed):

src/module.indexer/model/dftx/set.future.swap.ts
```diff
@@ -0,0 +1,51 @@
...
+    super()
+  }
+
+  async indexTransaction (block: RawBlock, transaction: DfTxTransaction<SetFutureSwap>): Promise<void> {
+    const txid = transaction.txn.txid
+    const data = transaction.dftx.data
+
+    await this.futureSwapMapper.put({
+      id: txid,
+      key: fromScript(data.owner, this.network)!.address,
```
fuxingloh, line 30: This will fail when a script cannot be parsed into one of the 3 human-readable addresses. E.g. multi-sig scripts.

You should sort as hex instead, e.g. `data.owner.toString('hex')`

https://github.com/JellyfishSDK/whale/blob/5f805aa8cdf1da3c02dcf76d0c592bb5ee136d31/src/module.indexer/model/dftx/place.auction.bid.ts#L29

fuxingloh (changes requested):

jingyi2811 (approved):

cwkang1998 (dismissed):

fuxingloh:
Targetting this for `0.35.x`

src/module.api/address.controller.ts
```diff
@@ -186,6 +190,28 @@ export class AddressController {
...
+  async listFutureSwap (
+    @Param('address') address: string,
+      @Param('height', ParseIntPipe) height: number,
+      @Query() query: PaginationQuery
+  ): Promise<ApiPagedResponse<FutureSwap>> {
+    const size = query.size > 30 ? 30 : query.size
+    const nextSettleBlock = await this.rpcClient.oracle.getFutureSwapBlock()
+    // mapper is sorted DESC
+    const lt = query.next ?? `${HexEncoder.encodeHeight(nextSettleBlock)}-${'f'.repeat(64)}`
+    const gt = `${HexEncoder.encodeHeight(height)}-${'0'.repeat(64)}`
```
fuxingloh, line 204: would it be better to get gt from something like 
`gt = nextSettleBlock - get future block interval?`

> get future block interval, I believe is already set in gov variable

canonbrother: huh?? the `nextSettleBlock` here is `future block interval` :sweat_smile: 
so output will just return the data within the settlement period

src/module.indexer/model/dftx/set.future.swap.ts
```diff
@@ -0,0 +1,51 @@
...
+    await this.futureSwapMapper.put({
+      id: txid,
+      key: toBuffer(data.owner.stack).toString('hex'),
+      sort: `${HexEncoder.encodeHeight(block.height)}-${txid}`,
+      source: {
+        token: data.source.token,
+        amount: data.source.amount.toString()
+      },
+      destination: data.destination,
+      withdraw: data.withdraw,
```
ivan-zynesis, line 37: They are full history of FutureSwap related actions, we have no intention to do any aggregation right?

canonbrother: hmm.. yea.. this is more historical-like data... 
understand the aggregation is more user read friendly.. perhaps that one for another api `:addr/future/swaps/:height/aggregation`

src/module.api/address.controller.ts
canonbrother: solved at https://github.com/JellyfishSDK/whale/pull/940/commits/f53b8b9a22e47716427e6b08e59eb41d677ee337

src/module.indexer/model/dftx/set.future.swap.ts
ivan-zynesis: Yea, trying to understand the API motive. Was thinking it could be meant to solve performance issue by not exposing the [rpc](https://github.com/DeFiCh/jellyfish/blob/d8a9f0e11e7cce0f1ab1815e9f5d88967e41aea1/packages/jellyfish-api-core/src/category/account.ts#L432)

fuxingloh: Yeah, this would be more like showing the history rather than aggregation.

fuxingloh:
We're finishing up https://github.com/JellyfishSDK/jellyfish/issues/580

This PR needs to be migrated to https://github.com/JellyfishSDK/jellyfish/tree/main/apps/whale-api

PR #975, opened 2022-05-20
feat(module.defid): Detect tip stale by height

### LINKED ISSUE #954: DeFiDProbeIndicator to include stale tip detection

`DeFiDProbeIndicator.ts` to include a stale tip detection in probes for health check.

https://github.com/JellyfishSDK/whale/blob/53dd32854167fb53bc38f5a3ef6fee027fe5e62f/src/module.defid/defid.probes.ts#L26-L57

While checking the connection count is enough 99% of the time, for the remaining 1% of the time a stale tip can still happen with connected nodes. We should detect stale tips after 90 minutes by checking if blocks came in? 

/area module-defid
/triage accepted
/assign @canonbrother

canonbrother:

Fixes #954


src/module.defid/defid.probes.ts
```diff
@@ -5,6 +5,8 @@ import { blockchain as bc } from '@defichain/jellyfish-api-core'
 
 @Injectable()
 export class DeFiDProbeIndicator extends ProbeIndicator {
+  lastBlock: bc.Block<bc.Transaction> | undefined
```
fuxingloh, line 8: Maybe we don't need to know the last block if we pull the data from the chain?

I am thinking also, that maybe we can pull it from model.probes.ts instead since it will ensure it's more end to end.

src/module.defid/defid.probes.ts
```diff
@@ -53,6 +60,19 @@ export class DeFiDProbeIndicator extends ProbeIndicator {
...
 
+    const highest = await this.modelProbe.block.getHighest()
+    if (highest === undefined) {
+      return this.withDead('defid', 'model.probe is not ready yet', details)
+    }
+
+    if (now() - highest.time >= 90 * 60 && highest.height >= block.height) {
+      return this.withDead('defid', 'chain is stale', details)
+    }
+
```
fuxingloh, line 71: I think should move the entire logic to `model.probes.ts` since it makes sense E2E we're checking Model for the reliability instead of defid now?

src/module.model/_model.probes.ts
```diff
@@ -32,13 +32,16 @@ export class ModelProbeIndicator extends ProbeIndicator {
    * - unable to get the latest block
    * - synced blocks are undefined
    * - synced blocks are more than 2 blocks behind
+   * - synced highest block height is still more than defid after 90 mins
```
fuxingloh, line 35:
```suggestion
   * - synced latest block height is not more than 90 mins old
```

src/module.model/_model.probes.ts
```diff
@@ -59,6 +62,14 @@ export class ModelProbeIndicator extends ProbeIndicator {
       return this.withDead('model', 'synced blocks are more than 2 blocks behind', details)
     }
 
+    if (now() - highest.time >= 90 * 60 && index >= defid) {
```
fuxingloh, line 65:
```suggestion
    if (index === defid && now() - highest.time >= 90 * 60) {
```

Since index defid can experience a rollback, let's make it such that index is `always === to defid` to make sure the stale condition is only checked if `Model === DeFid`.

fuxingloh (changes requested):
Looks good with minor condition change request

src/module.api/actuator.e2e.ts
```diff
@@ -200,4 +200,44 @@ describe('with peers', () => {
       status: 'ok'
     })
   })
+
+  it('/_actuator/probes/readiness stale', async () => {
+    // Time travel 100 mins
+    Date.now = jest.fn(() => new Date().getTime() + 6000_000)
```
fuxingloh, line 206: I don't think this way of injection can be mock cleared? Might it affect other tests? Better to inject it into the `ModelProbeIndicator` class via nestjs. Otherwise this testing of jest mock might affect other tests down the road.

src/module.api/actuator.e2e.ts
```diff
@@ -200,4 +200,46 @@ describe('with peers', () => {
...
+          headers: expect.any(Number),
+          initialBlockDownload: false,
+          peers: 1,
+          status: 'up'
+        }
+      },
+      status: 'error'
+    })
+
+    mockTime.mockRestore()
```
canonbrother, line 243: safety clear - do `mockRestore` before end of test

src/module.api/actuator.e2e.ts
canonbrother: yea you're right.. added `mockRestore`

fuxingloh (dismissed):

src/module.model/_model.probes.ts
```diff
@@ -59,6 +62,15 @@ export class ModelProbeIndicator extends ProbeIndicator {
       return this.withDead('model', 'synced blocks are more than 2 blocks behind', details)
     }
 
+    // index defid can experience rollback, so make sure the condition is only checked if `Model == DeFid`
+    if (index === defid && now() - highest.time >= 90 * 60) {
```
eli-lim, line 66:
```suggestion
    if (index === defid && secondsSince(highest.time) >= 90 * 60) {
```

Minor suggestion for readability(?) haha

eli-lim (dismissed):
LGTM! just some minor comments

src/module.api/actuator.e2e.ts
```diff
@@ -200,4 +200,46 @@ describe('with peers', () => {
       status: 'ok'
     })
   })
+
+  it('/_actuator/probes/readiness stale', async () => {
+    // Time travel 100 mins
+    const mockTime = jest.spyOn(Date, 'now').mockImplementation(() => new Date().getTime() + 6000_000)
```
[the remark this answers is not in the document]
jingyi2811: I was thinking the same too

jingyi2811 (approved):

